PR #94, opened 2013-09-20
Fixed rotation center coordinates.

sdussaut:
When setting a rotation, SVG.js calculates the origin point of the transform based on the object cx and cy. I corrected the test which was testing null while browsers send undefined and therefore coordinates were set to NaN.


Fuzzyma:
@sdussaut -  is this still an issue

PR #125, opened 2013-12-18
defs and bbox fixes

lent:
small fixes that were necessary for more extensive svg transformations.
Edit: The bbox change is not really necessary. For curiosity: Why do you  not use element.getTransformToElement(svg) and SVGPoint.matrixTransform() to get transformations?


Fuzzyma:
@lent - is anything of this still relevant / not implemented?

PR #180, opened 2014-05-03
Add support for removing SVG root by remove method

erykpiast:
If <svg> element is attached to HTML element (which is, I suppose, common use case), you can remove it by calling `remove` method of object returned by SVG() function. The same like you remove SVG elements from SVG document.


Fuzzyma:
Fixed with e5b321e1c49928d5e7fceeeedec7fde91aa5e0f0

PR #272, opened 2014-10-09
Remove buggy 2nd module export

marrs:
You are defining your module twice which is causing issues for projects
using your library with RequireJS. See https://github.com/jrburke/requirejs/issues/1216 for details.


zachrbrown:
Closed via #312

PR #274, opened 2014-10-17
Add support for event namespaces

mruoss:
This will add support for namespaces on events known from jQuery (see http://css-tricks.com/namespaced-events-jquery/). My problem was that I was using the draggable plugin and I has problems registering a mousedown event because draggable also registered a mousedown event on the same element and unregistered it when I called fixed() on that element. My own event handler was removed, too. With namespaces, I can register and unregister "my own" events.

Example:
```javascript
rect.on('mousedown.myPlugin', function(event) {
     // do the work
});
```
```javascript
rect.off('mousedown.myPlugin');
```

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -50,7 +50,7 @@ SVG.on = function(node, event, listener) {
   SVG.listeners[node][event][listener] = l
 
   // add listener
-  node.addEventListener(event, l, false)
+  node.addEventListener(event.split('.')[0], l, false)
 }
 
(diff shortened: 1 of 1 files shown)


Fuzzyma:
I am planning on implementing this.
Your solution grasp short a bit (as @dustinmoorenet mentioned)

Fuzzyma:
closed with 0d1e85f6e3953461808c70366528c2ac3ce7903b

PR #307, opened 2015-01-05
The use element now can point to an external file

GSchutz:
Just added a second argument to the method, the SVG file path/url.


Fuzzyma:
Please add specs. I will accept it then.

GSchutz:
@Fuzzyma 
Sorry for the delay, specs created and update the Readme.

Fuzzyma:
meeh - its not mergeable anymore :/. Can you do a rebase to the current working state?

GSchutz:
@Fuzzyma 
I was in doubt on doing this in the right way, but when I discard the `dist` and merged there was no conflict, and then I generate the dist and commit again.

Hope is the right move.

Fuzzyma:
Actually I have no idea :D. I am no git pro either.
I saw that you merged the current master branch into you branch. In my opinion it should be the opposite. You merge _your_ branch into the _master_ and not the other way round.
When I create a PR I usually create a new feature-branch where I put my changes so that I can pull the newest master every time and have my changes in a different branch. But I am sure there are 100 other ways doing this. 

I dont know what will happen when I merge this now. Thats why I have to doublecheck that.
Maybe its the best possible solution when I just do the changes on the code and mention you in the changelog instead (?). That would save me from merging something without knowing what will happen

GSchutz:
If you look at the files changed, it just have the changes that I made, not removing any additional code, besides that, all tests are ok to. If you really are in doubt, you can rollback after the merge.

PR #309, opened 2015-01-07
Removed superfluous UMD wrapper/AMD check that causes issues with Requir...

bcbweb:
...eJS

### DIFF
--- a/dist/svg.js
+++ b/dist/svg.js
@@ -12,68 +12,68 @@
   var SVG = this.SVG = function(element) {
     if (SVG.supported) {
       element = new SVG.Doc(element)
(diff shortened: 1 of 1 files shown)


stephanebachelier:
@bcbweb your PR is not correct as you have modified the dist file and not the src file.
I would add that if you touch on the AMD loader you should not modify the whitespace or create a commit to change all the whitespaces.

And last why would you want to remove the UMD wrapper ?

bcbweb:
@stephanebachelier Apologies, the request was an accident (I meant to merge it into the master on my fork for testing purposes)

PR #318, opened 2015-01-30
Call window.cancelAnimationFrame to break out of fx.render loop

zachrbrown:
Addresses the severe zombie animation loop documented at https://github.com/wout/svg.js/issues/316


MathieuLoutre:
Nice! Good job including the polyfill too. I don't know the PR policy, but you might want to make that change here: https://github.com/wout/svg.js/blob/master/src/fx.js

Fuzzyma:
Thank you for that PR!
Would you mind adding spec-examples for this PR (so we have a test for that behavior later)

PR #351, opened 2015-06-11
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3


Fuzzyma:
" Intended to be used in the future when Bower gets a real registry" 

So I would add it in the future again? Well then I can leave it as it is. 
I find it very useful to see the actual version in the Bower file.

PR #372, opened 2015-08-17
Transform lists in IE11

rayglover:
IE11 doesn't use a comma to delimit transform-list elements. This PR intends to make parsing matrix transforms more in-line with the W3C spec.


eXxi:
Thank you very much :) Moving a group in IE11 is not possible without this fix.

Fuzzyma:
If you could run the gulp one time and add this to your PR i will accept it

rayglover:
Merged and rebuilt (still v2.0.5)

bubenshchykov:
@rayglover thanks man you rock!
@Fuzzyma looking forward to have it on bower :)

PR #407, opened 2015-11-15
Reorganize and clarify documentation

snowyplover:
I see what you mean about the issue of user coordinates being more general than the way I initially identified it.

However, in order to make a sensible place to put the warning I wanted to add, I needed to reorganize some of the documentation sections. 

I have replaced the 'Manipulating elements', 'Attribute references', and 'Inserting elements' sections with 4 new sections:

* Attributes and styles
* Classes
* Transforming elements
* Altering the element tree

I have moved the method descriptions around as follows:

In 'Attributes and styles':
* `attr, style, reference, hide, show, visible`

In 'Classes':
* `classes, hasClass, addClass, removeClass, toggleClass`

In 'Transforming elements':
* `transform, size, width, height, move, x, y, center, cx, cy, dmove, dx, dy`

In 'Altering the element tree':
* `clone, remove, replace, add, put, addTo, putIn, toParent, toDoc, ungroup/flatten`

I've placed a note at the beginning of 'Transforming elements' that is based on the information that was originally in the description of the `move` method, and which includes a warning about using user coordinates.

I also ended up editing many of the method descriptions in the sections I reorganized.

Of course, I removed the warnings from `center`, `cx`, and `cy`. My notes on `SVG.invent()` I left in, though.

It seems very awkward to have the documentation in one large file like this. Plus, the diff seems too messy to show the changes I made clearly. So I'm not sure I'm using Github very effectively here. I hope what I'm doing is actually helpful, but please let me know if I'm just making trouble for you.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3701,7 +3701,7 @@ It should be emphasized that in the configuration object passed to `SVG.invent()
 - `construct` does not supply constructors, but rather methods that are likely to *call* constructors;
 - `create` specifies the constructor for the type you are defining, and is not analogous to `Object.create()`.
 
-When defining specialized svg elements (such as SVG.Rounded in the example above), the function specified by `create` needs to do all the work of adding the element to the DOM for the svg document and connecting the DOM node to the SVG.js interface. All this is done automatically when the value of `create` is a string identifying an element type. If needed, see the source for a sense of how to do it explicitly.
+When defining specialized svg elements (such as `SVG.Rounded` in the example above), the function specified by `create` needs to do all the work of adding the element to the DOM for the svg document and connecting the DOM node to the SVG.js interface. All this is done automatically when the value of `create` is a string identifying an element type. If needed, see the source for a sense of how to do it explicitly.
 
 Though the defaults are geared toward creating svg elements for the SVG.js framework, `SVG.invent()` can be used as a generalized function for defining types in Javascript. When used in this more general way, the function supplied as a value for `create` should be written as an ordinary JS constructor. (Indeed, the function is simply returned as the constructor for your newly defined type.)
 
--- a/README.md
+++ b/README.md
@@ -1609,7 +1609,7 @@ rect.putIn(group) //-> returns group
 __`returns`: `element`__
 
 ### toParent()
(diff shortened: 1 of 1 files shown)


Fuzzyma:
Thanks for this bunch of work you put into this.
I think the new structure is logical and well thought trough. My only remark is to clarify the use of "tranformation". When you talk about "transforming elements" in svg it generally means the use of the transform-property (rotation, scale...). I think that could lead to confusion since this is used for move, cx... here, too. I admit, that I have no other naming-idea for this section but I would feel better when the headline would be more... general (?).

snowyplover:
Yes, "Transforming elements" isn't the best label. I was concerned about that too, but couldn't come up with anything better at the time.

And I agree it is most likely better to keep methods that affect the transform matrix conceptually separate from the more direct, attribute-based methods like `move()`. 

I also realized that I should have incorporated the "Syntax sugar" methods into the new categories.

The new commit:
* Renames "Transforming elements" to "Size and position", and moves the `radius()` method to that category. (I also edited `radius`.)
* Moves `fill()`, `stroke()`, and `opacity()` to "Attributes and styles".
* Groups `rotate()`, `skew()`, `scale()`, and `translate()` with `transform()` under "Geometry". Does that make sense? If not, then it would be easy to give those methods their own category, "Transforms".

I'm also not really sure about the term "element tree". The point is to make clear that the structure that is referred to is the hybrid DOM/svg.js structure. In other contexts, one could just use "DOM" or "document tree", but that doesn't seem to work here. "Element tree" isn't standard, so at least it calls attention to itself, but it's not particularly clear. I suspect this is the sort of term that needs to be explicitly established as a definition that will be used by the documentation, and then used consistently throughout.

Fuzzyma:
Jej, awesome! I vote for the "Transforms"-Category :).
I think I would still call it the dom tree. So DOM Manipulation didnt sound wrong to me.
After all this are shortcuts to alter the DOM. You ofc always get an SVG-Element back and no DOM-Element but I dont think that anyone will wonder about this.

snowyplover:
I've created the "Transforms" category for the five transform matrix methods.

I also renamed "Altering the element tree" to "Document tree manipulations", and edited the note on "Size and position" to reflect the fact that `transform()` is no longer part of that category.

PR #412, opened 2015-11-21
Fixed support amd

chuve:
Added wrap factory function for correct work with AMD.


Fuzzyma:
Can you please explain how this works and how you would use it now with amd? I myself don't use it so I would appreciate some information here :)

chuve:
Sure! 
I use svg.js through requirejs (http://requirejs.org/)
```javascript
require(['svg'], function(SVG) {
          var draw = SVG('drawing').size(300, 300)
          var rect = draw.rect(100, 100).attr({ fill: '#f06' });
});
```
Example of this code here: http://jsbin.com/quduzeyeki/2/edit?html,js,console,output

With current last version of svg.js, I have error: 
```javascript
TypeError: SVG is not a function(…)
```
When requirejs define module of svg.js by factory function it wait for dependencies.
Take a look at factory function:
```javascript
function(window, document) {}
```
For require.js function parameters mean dependencies.
Have a look this example: http://requirejs.org/docs/api.html#defdep
At first thought correct code for define svg.js module is:
```javascript
define([root, root.document], factory)
```
But of course... window and window.document are not depending modules ;)

So I made new function with preset parameters for factory.
```javascript
var amdFactoryFunction = function() {
  return factory.apply(factory, [root, root.document]);
};
```
And now it works! :)

Fuzzyma:
Well the window object is more or less a dependency of this lib. It doenst work without one.
So on serverside you have to apply our own window.

How would that work in amd?

chuve:
We get it from root variable. In AMD it is window global object.

Fuzzyma:
Also on serverside? (nodejs)

chuve:
On serverside root is "global" node.js object and "window" and "document" takes from factory function parameters.

Have a look at UMD Wrapper:
```javascript
(function(root, factory) {
  /* === this code run in an AMD environment (1) === */
  if (typeof define === 'function' && define.amd) {
    var amdFactoryFunction = function() {
      return factory.apply(factory, [root, root.document]);
    };
    define(amdFactoryFunction);
  /* === end (1) === */
  /* === this code run in node.js environment (CommonJS) (2) === */
  } else if (typeof exports === 'object') {
    module.exports = root.document ? factory(root, root.document) : function(w){ return factory(w, w.document) };
  /* === end (2) === */
  } else {
  /* === this code run in global window scope <script src=""></script> (3) === */
    root.SVG = factory(root, root.document);
  }
  /* === end === (3) */
}(typeof window !== "undefined" ? window : this, function(window, document) {
```

Fuzzyma:
yes, nodejs is using commonjs to load its modules as default. But AMD was designed to unify the loading in browser and nodejs and therefore amd can be used on serverside, too (thats more or less a question :D)

Fuzzyma:
@chuve can you please update the PR. I forgot to merge it before my last commits *grmpf*. Sorry!

Fuzzyma:
I added this to the latest release. Thank you!

Issue #425, opened 2015-12-08
rbox() returns NaN in IE for x & y properties


jcvandan:
I noticed this happening then when I dug into the problem, it is because IE does not implement the `window.scrollY` and `window.scrollX` properties which are used as an offset on the x & y properties (see line 1829 of dist file). This is easily resolved by having an || and using the respective properties that IE provides:

```javascript
// offset by window scroll position, because getBoundingClientRect changes when window is scrolled
this.x += window.scrollX || window.pageXOffset
this.y += window.scrollY || window.pageYOffset
```

I will happily fix this issue and submit a pull request if it is agreed that this is a reasonable fix.

Fuzzyma:
Yes thank you! Isn't pageYOffset the standard anyway? Maybe we need only
this. May you check this?

Dan Ormisher <notifications@github.com> schrieb am Di., 08.12.2015, 14:17:

jcvandan:
Hey, it [looks like](http://www.w3schools.com/jsref/prop_win_pagexoffset.asp) `pageXOffset` is supported across the board and from IE 9+. I believe svgjs only supports IE 9+ so I assume this is fine. I have just done a quick test and compared the values of `scrollY` and `pageOffsetY` in the latest Chrome and Firefox and they both return the same value for each property.

I will do this now and submit a pull request.

Fuzzyma:
Yes thank you

Dan Ormisher <notifications@github.com> schrieb am Di., 08.12.2015, 14:41:

Fuzzyma:
Closed with the PR

PR #426, opened 2015-12-08
Replaced usage of window.scrollX with window.pageXOffset

jcvandan:
This is the fix discussed in issue #425


Fuzzyma:
Thank you!

PR #440, opened 2016-01-15
Update boxes.js

Samuel3:
In Firefox I had this problem that these commands goes straight into infinity loop. Because the cloned object could also not calculate bbox so it made a new cloned one. This new clone could also not calculate the bbox and so on.

var clone = element.clone().addTo(SVG.parser.draw)
          box = clone.bbox()
          clone.remove()

### DIFF
--- a/src/boxes.js
+++ b/src/boxes.js
@@ -10,17 +10,11 @@ SVG.BBox = SVG.invent({
         // find native bbox
         box = element.node.getBBox()
       } catch(e) {
-        if(element instanceof SVG.Shape){
-          var clone = element.clone().addTo(SVG.parser.draw)
-          box = clone.bbox()
-          clone.remove()
-        }else{
-          box = {
-            x:      element.node.clientLeft
-          , y:      element.node.clientTop
-          , width:  element.node.clientWidth
-          , height: element.node.clientHeight
-          }
+        box = {
+          x:      element.node.clientLeft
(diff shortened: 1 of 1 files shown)


Fuzzyma:
Well - just removing the lines is not a solution to the problem. I fixed this infinite loop several times but it kinda always comes back. However these lines are important and need to stay. Please post your code where this error occurs!

Samuel3:
The problem I have is in a web portal with an application that runs in a iFrame. My code is the following:
From the iFrame I add a Image to an SVG Group in the main window. I know this is a bad situation with  this iFrame construction, but I can not change this. Shrinked my code is the following: parent.SVG.group().image("http://srvutdemo102/module/webcodi/../../../../utjsc_webgen/user/script/globespotter/img/globespotter_marker_symbol.png",540, 442, 130, 250);
I tried to create a simple test page with this failure, but the failure is hard to recreate outside of the mentioned environment. 

Beside of that I find it difficult to create a recursive with no stop criterion. Maybe something with check on callee.caller could work.

Fuzzyma:
Thanks for the info. Could I please have your exact structure and more code?

Samuel3:
Well it's a little bit difficult to provide more of code. The situation is that there is software from 3 different vendors which works toghether. It's hard for me to list the exact code. We talk about 100k LOC. But I try to describe it bether.

Issue #438, opened 2016-01-14
Line disappears when moved outside of body's bound. 


raphaelduartepinheiro:
Hello. Today while I was testing some movements with svg.js I've found something that I don't know if it's a bug or an expected behavior.

If I move a set of lines to somewhere outside of body's bound, and try to come back with it, one line disappear.

Example: 

https://jsfiddle.net/t13tdkr0/

Use scroll of the mouse to simulate: 1 scroll up, 2 scrolls down.

If I apply ``position: absolute`` to the line element, the problem doesn't occurs.

My question: is this behavior correct? I wanted to "hide" the elements and show, going out of screen, but the vector layout seems to break.

Appreciate any reply.

Fuzzyma:
I cant reproduce your issue. The jsfiddle just does nothing :D

raphaelduartepinheiro:
Sorry if I wasn't clear... Did you tried to scroll with the mouse wheel? 

Using Chrome here.

Fuzzyma:
Maybe the error shown is the problem:
```
[...]

PR #444, opened 2016-01-17
Fix regex for isNumber

snowyplover:
Description of necessary functionality for a regex that matches Javascript numbers found in Eloquent Javascript, (Ch 9, last exercise). See: http://eloquentjavascript.net/09_regexp.html#h_izldJoT3uv

Probably solves issue #438.

### DIFF
--- a/src/regex.js
+++ b/src/regex.js
@@ -34,7 +34,7 @@ SVG.regex = {
 , isBlank:          /^(\s+)?$/
   
   // Test for numeric string
-, isNumber:         /^-?[\d\.]+$/
+, isNumber:         /^[+-]?(\d+(\.\d*)?|\.\d+)(e[+-]?\d+)?$/i
 
   // Test for percent value
 , isPercent:        /^-?[\d\.]+%$/


snowyplover:
I suggest adding the tests from [Eloquent Javascript](http://eloquentjavascript.net/09_regexp.html#h_izldJoT3uv) to the test suite. I've never used Jasmine, so I wasn't quite sure how to do this, but I could figure it out if necessary.

Fuzzyma:
Adding tests for all the regex is a very good idea.
Jasmin has an easy syntax. For example see this one:

```javascript
// descriptions can be nested to describe subgroups
// everything is just a function
describe('PointArray', function () {
	// All `it()`s belong to one `describe`
	it('parses a string to a point array', function() {
		var array = new SVG.PointArray('0,1 -.05,7.95 1000.0001,-200.222')

		// the expectation is the test. some value has to be something (or not to be) to pass the test
		// you can have multiple `expect()`s in one `it()`
		expect(array.valueOf()).toEqual([[0, 1], [-0.05, 7.95], [1000.0001, -200.222]])
	})
	it('parses a points array correctly to string', function() {
		var array = new SVG.PointArray([[0,.15], [-100,-3.141592654], [50,100]])

		expect(array + '').toBe('0,0.15 -100,-3.141592654 50,100')
	})
})
```

PR #447, opened 2016-01-22
Uncaught TypeError: Cannot read property 'node' of null

assembler:
I've encountered situation where SVG.adopt (line 176) returns null, which leads to `Uncaught TypeError: Cannot read property 'node' of null`. Checking whether parent exists fixes it.

### DIFF
--- a/src/element.js
+++ b/src/element.js
@@ -171,7 +171,7 @@ SVG.Element = SVG.invent({
       if(!type) return parent
 
       // loop trough ancestors if type is given
-      while(parent.node instanceof SVGElement){
+      while(parent && parent.node instanceof SVGElement){
         if(typeof type === 'string' ? parent.matches(type) : parent instanceof type) return parent
         parent = SVG.adopt(parent.node.parentNode)
       }


Fuzzyma:
In which situation does this happen?
Its very unlikely that an element which has at least one existing parent (which is checked before) doenst has more parents.
The only possibility where I imagine this could happen is when you hit the html-element which is not possible duo to the `instanceof SVG.Element` check. So iam very curious about your code!

assembler:
I have tried to reproduce this issue outside of my project, but with no luck. The only time it happens is when i want to place image on canvas, but only on initial page load. Everything renders fine even though the error is thrown. Perhaps it has to do something with redrawing of "canvas" (whenever something is added to the graph generated by svg.js, the canvas is cleared, and it is redrawn from scratch). Maybe its getting redrawn after image load is initialised, but onload callback fires after redrawing, so it could relate to detached elements.. I'm not sure.

Fuzzyma:
I am not getting off that feeling that this fix would only solve the effect and not the cause. However it could be somehow related to the onload thing of images. I thought I solves this some time ago... 
The onload function is executed when the image is not in the dom anymore because you redraw the canvas. So it doesn't find the parent... 

Looks like I have to accept this PR ^^

Fuzzyma:
Please update the changelog (2.3.0) and Run gulp to reflect the change in the dist files.

PR #448, opened 2016-01-26
fix Inside function bbox to tbox

gmixo:
It seems inside function should use tbox instead of bbox because after move() bbox still returns {x:0, y:0} but object actually change his position.

I'm can't understand actual usability of current version of bbox, it seems that prev version of bbox that currently has name tbox is what I'm needed

### DIFF
--- a/src/element.js
+++ b/src/element.js
@@ -97,7 +97,7 @@ SVG.Element = SVG.invent({
     }
     // Checks whether the given point inside the bounding box of the element
   , inside: function(x, y) {
-      var box = this.bbox()
+      var box = this.tbox()
 
       return x > box.x
           && y > box.y


Fuzzyma:
Can I please see your code where you got that problem?

gmixo:
Main difference is that I use translate() method

var rect = draw.rect(100, 100).translate(50, 50)
in that case rect.bbox() != rect.tbox()

rect.inside(25, 30) //-> returns true
rect.inside(60, 70) //-> returns true
rect.inside(110, 110) //-> returns false

Fuzzyma:
Yes exactly. The inside function checks for points inside this rectangle in the coordinate system of the rectangle. If you use translate you dont move the rectangle but the coordinate system.
This is intended because if you would allow transformations you also would have to check for skew and rotation which is not easily possible.
Long story short: Your use is not intended and wont work

gmixo:
ok thanks for answer

gmixo:
one more silly question if I move a group it bbox is also don't changes.
wy groups won't moves?

Fuzzyma:
Because a group can't be moved with x and you attributes. It gets his position from its content. I am not completely sure if bbox changes when you move the content. However if not you could use the unified box from its content. The only way a group can be moved is with translation and then the thing above strikes again :). There is a special box called gbox() which gives the translated position of a group if you need it.

gmixo:
thanks again)

Issue #409, opened 2015-11-16
SvgPathSeg is removed in Chrome 48


jurebajt:
Hi!

According to https://www.chromestatus.com/features#pathseg SVGPathSeg interface was a part of svg1.1 and will be removed. This brakes the drawing of paths in Chrome Canary (Version 48.0.2564.0 canary (64-bit)) using SVG.js version 2.2.0.

```
Uncaught TypeError: Cannot read property 'numberOfItems' of undefined
```

Call stack:
```
SVG.extend.parse (SVG.js:669)
SVG.Array (SVG.js:387)
SVG.PathArray (SVG.js:553)
SVG.Path.SVG.invent.extend.plot (SVG.js:3362)
SVG.Path.SVG.invent.construct.path (SVG.js:3397)
```

danburzo:
:+1: I was about to write this one up as well
There's a polyfill that might be helpful: https://github.com/progers/pathseg
More discussion, for context: https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/EDC3cBg9mCU

Fuzzyma:
>**Compatibility Risk**
>Low, this feature is relatively obscure and didn’t gain much usage so the SVG WG decided it was best to remove it before usage grows.

Well...

> Can you please confirm for me that after your removal the only way to access this information in Chrome will be for us to parse the `d` attribute ourselves?

> Until the new API [1] matures and is implemented - yes that is correct

Is this a joke? Better lets remove a feature because a better feature *might* be coming in the future.

Sometimes I don't get these people...

However a big thanks for posting this!
I will try to fix this in 2.2.1

Fuzzyma:
In my tests the new code was 100% faster - and its even less code ^^

jurebajt:
Awesome!

pjfsilva:
I know this is a hard request but I need to be stuck on svg.js 1.1.0 because of plugin dependencies so...

Is there any possibility we can backport this fix to 1.1.0 branch?

ibrierley:
You can always include the polyfill listed above I assume.

Fuzzyma:
yes thats right. just include the polyfill and it will work

PR #462, opened 2016-03-02
Backports the fix to #409 from the master branch to v1.x branch.

pjfsilva:
Backports commit cf783fca1213d26b50bc07af0ca7e2b0f6385401
Backports the fix for #409 "SvgPathSeg is removed in Chrome 48".
This change fixes 53 test failures on Chrome 48+ (error: 'TypeError: Cannot read property 'numberOfItems' of undefined'.


Fuzzyma:
did you run the tests in chrome? Also I am pretty sure you missed 2 commits which fixes issues with this fix (dd7cc2f5104d488b0ba85366e14fae57edb767b7 and dd2ad7824d546242468992436d9a631a60294e28)

pjfsilva:
Yes, ran tests on Chrome 48 and 51 canary. All 53 tests that were failing because of SvgPathSeg being removed on Chrome 48 (error: 'TypeError: Cannot read property 'numberOfItems' of undefined'.) now run correctly.

I wasn't aware of any further commits related to this, since all tests are now working. I'm not sure if those are easy to backport to v1.x, do you want me to try to backport also those or we do different PRs?

pjfsilva:
Updated PR to include dd7cc2f and dd2ad78.
patharray.js is now synchronized to HEAD on master branch

Fuzzyma:
I wasnt speaking from the tests *before* the fix but *after*. Are all tests ok after applying this patch?
If so I will merge it immediately

pjfsilva:
OK, got it.

There is 1 test failing both *before* and *after* on Chrome 48 and 51 (and IE and FF) but it doesn't seem to be related to this PR. I believe that test is not correct but I don't have enough background to understand if the spec is wrong or there is a real problem.

Test failing:
`Regex testers isEvent is false without a namespace`

That test doesn't exist anymore on master and neither the isEvent regex. Other than this one, all tests are OK.

Fuzzyma:
Published under 1.1.1

PR #470, opened 2016-04-01
added svg.js.d.ts

rdfriedl:
seeing as typescript is starting to get more popular.
and with angular2 in beta, i thought it would be a good time to write the typescript definitions for this library.
i dont own all the credit for the definitions, i started with the definitions i found on http://definitelytyped.org/ and expanded on that. [here is the original](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/svgjs/svgjs.d.ts)

anyways feel free to review / change it since im sure i miss a few things :D

also i did not write the definitions for SVG.FX and the animations in general, i ran out of time :D

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
   },
   "github": "https://github.com/wout/svg.js",
   "license": "MIT",
+  "typings": "./svg.js.d.ts",
   "devDependencies": {
     "del": "^1.2.0",
     "gulp": "^3.8.6",
--- a/svg.js.d.ts
+++ b/svg.js.d.ts
@@ -1,0 +1,893 @@
+// todo add SVG.FX
+declare namespace svgjs {
+    export interface Library {
+        (selector: string): Doc;
+        (domElement: HTMLElement): Doc;
+        ns: string;
+        xmlns: string;
+        xlink: string;
+        svgjs: string;
+        supported: boolean;
+
+        did: number;
+        eid(name: string): string;
+
+        create(name: string): any;
+        extend(parent: Object, obj: Object): void;
+        invent(config: Object): any;
+        atopt(node: HTMLElement): Element;
+    }
+    interface LinkedHTMLElement extends HTMLElement {
+        instance: Element;
+    }
+
+    // arrange.js
+    interface Element {
+        front(): this;
+        back(): this;
+        forward(): this;
+        backward(): this;
+
+        siblings(): Element[];
+        position(): number;
+        next(): Element;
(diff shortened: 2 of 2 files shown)


rdfriedl:
i also just finished one for svg.filter.js so if you want i can create a PR for that one to.

the reason i made this one and the one for svg.filter.js is because im currently working on a typescript project and i need to use svg.js (because its very useful) :D

Fuzzyma:
I guess thats a stupid question but I dont know typescript: so what is that for?

rdfriedl:
in simplest form they are files that explain to typescript what each function / property in the library dose.
that way it can know now to handle type checking for when the user try to use the library

so for example if it write:
```typescript
var canvas:SVG.Doc = new SVG('#some-svg');
```
_the `:SVG.Doc` tells typescripts that my var is going to contain a SVG.Doc class_

but by default typescript dose not know how to handle other libraries.
so you write the definition files to tell is that `SVG()` is a functions and that `SVG.Doc` is a class that has certain properties and methods.

if you want to read more about it the documentation is  [here](https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/Writing%20Declaration%20Files.md)

Fuzzyma:
Thanks for you work! I think we can merge this!

Linkviii:
This doesn't seem to be published anywhere? Typings has an old (the original source?) definition from DefinitelyTyped.

Fuzzyma:
Sorry, I dont get your point. As far as I can tell the typings got corrected before beeing merged here

Linkviii:
```
$typings search svgjs
Viewing 2 of 2

NAME            SOURCE HOMEPAGE              DESCRIPTION VERSIONS UPDATED
svgjs           dt     http://www.svgjs.com/             1        2016-03-17T12:06:54.000Z
svgjs.draggable dt     http://www.svgjs.com/             1        2016-06-02T14:39:50.000Z
```
installed dt~svgjs and the definition does not match the one in this repository.

Fuzzyma:
I dont even know that you have to publish this typescript definitions as extra again. I am pretty sure http://definitelytyped.org/ published that.

Linkviii:
Searching for svgjs with Typings only yields the DefinitelyTyped definition is what I'm saying.

rdfriedl:
i dont think this version is on Typings (DefinitelyTyped)
i never put it up there because im not familiar with how DefinitelyTyped works, though feel free to put it up :D

PR #481, opened 2016-04-25
Fix groups animation

mkorganashvili:
When you use 'move' animation of the group element, and use transformation for only one coordinate (x or y) then it's resets other coordinate to 0. 
For example if your group is on coordinate 100, 100 and you call animate y(500), it goes to 0, 500.

This is caused because undo is made for both coordinates.

### DIFF
--- a/src/transform.js
+++ b/src/transform.js
@@ -263,7 +263,12 @@ SVG.Transformation = SVG.invent({
     }
 
   , undo: function(o){
+      for(var i = 0, len = this.arguments.length; i < len; ++i){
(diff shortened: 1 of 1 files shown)


Fuzzyma:
Is there any reason why this was never merged?

ramezrafla:
@wout 
This is indeed a bug (sadly, this is was my first animation as I was trying out this library, got me to almost walk away).

This PR fixes it (thanks @mkorganashvili), can we please merge (others will undoubtedly face this same issue). Thanks

PR #490, opened 2016-05-24
Changed document.contains to document.body.contains

snydesc:
The document object in IE does not have the contains function  as a method.  This can cause a stack overflow as you can get stuck in an endless try catch loop.   document.body.contains will work in IE, Firefox, and Chrome and will resolve this issue.

### DIFF
--- a/src/boxes.js
+++ b/src/boxes.js
@@ -9,7 +9,7 @@ SVG.BBox = SVG.invent({
       try {
 
         // the element is NOT in the dom, throw error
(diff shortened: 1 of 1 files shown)


Fuzzyma:
The problem with this notation is, that there is no `body` in svg-only documents.
However `document.documentElement` would work. I am not sure, if IE implements `contains` on the `html` node?

snydesc:
I updated my submission to use document.documentElement.

PR #504, opened 2016-06-23
Fix broken animations, if using polyfills for es6/7 proposals 

tkiss:
In this case the String.at() polyflll interferes with the check for `el.at`, resulting in animations trying to animate attributes not existing like "o" in case of "opacity"

### DIFF
--- a/src/fx.js
+++ b/src/fx.js
@@ -543,7 +543,7 @@ SVG.FX = SVG.invent({
       for(i in s.animations){
 
         at = [].concat(s.animations[i]).map(function(el){
-          return el.at ? el.at(s.ease(self.pos), self.pos) : el
+          return typeof el !== 'string' && el.at ? el.at(s.ease(self.pos), self.pos) : el
         })
 
         target[i].apply(target, at)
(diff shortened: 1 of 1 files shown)


Fuzzyma:
It would be awesome if you could provide a test case for this!

tkiss:
I finally managed to set up a test case for this:

[SVG.js Test on github](https://github.com/tkiss/svg.js-test)

Fuzzyma:
Mh you assume that webpack and jade is globbally installed. Can you chage that?

tkiss:
Usually it should work, I also haven't jade and webpack installed globally.
If it does not, I will try to figure out a workaround in the next days.

Fuzzyma:
I run your commands and it failed at multiple points (iam on a windows machine though...)

tkiss:
Ok, I run it on my windows machine (the test case was set up on OS X), also without webpack and jade globally, and ran into problems with `mkdir -p`. I fixed that and the chaining of the npm scripts.
I also included the /dist folder to the repo for a first access to the final output for testing.

Please let me know, if it now does work out :)

funklos:
@Fuzzyma will you merge this pr? i also ran into this issue and would rather use svg.js instead of a local fork

PR #527, opened 2016-10-04
Unittest

dotnetCarpenter:
I wanted to work on #525 by writing a failing test but to my surprise I couldn't run `npm test` and when I finally ran the jasmine test on my local server, I saw tests failing.

So instead of working on #525, I've updated the test suite so it can run on travis. I think that makes it more clear to everyone, when tests are failing. I also updated the dependencies and made sure that they work out-of-the-box (like gulp).

If you want it to run on travis, @wout, have to enable it. And you will want to update the badge in the README.md.

There is still some difference between running the tests via Karma and just opening the test suite in a browser. Not sure why yet, but I will try to iron it out in later commits.

But I hope the authors of the tests will take a look and see if any of their tests are failing.

The manual browser test has 5 failing tests (7 if tests are run random):
1.Element point() creates a point from screen coordinates transformed in the elements space `Expected -22 to be close to -21.`
2. Matrix inverse() inverses matrix `Expected 'matrix(0.5,0,0,0.20000000298023224,-2,-0.6000000238418579)' to be 'matrix(0.5,0,0,0.2,-2,-0.6)'.`
3. Text y() returns the value of y without an argument `Expected -4.470348358154297e-8 to be 0.`
4. PathArray size() resizes all points in a arc path `Expected 'M80 80A600 200 0 0 0 680 280L680 80Z ' to be 'M80 80A599.9998982747568 199.9999660915856 0 0 0 679.9998982747568 279.99996609158563L679.9998982747568 80Z '.`
5. FX at() sets the progress to the specified position `Expected 1475598310479 to be less than 1475598310479, 0.0001.`

Whereas the cli version has 13 failing tests:
1. PathArray size() resizes all points in a arc path FAILED
	Expected 'M80 80A600 200 0 0 0 680 280L680 80Z ' to be 'M80 80A599.9998982747568 199.9999660915856 0 0 0 679.9998982747568 279.99996609158563L679.9998982747568 80Z '.
	@spec/spec/array.js:77:7
2. Element svg() without an argument returns full raw svg when called on the main svg doc FAILED
[...]

PR #529, opened 2016-10-11
Pointarray - parse coordinates with Illustrator style space and or comma

### LINKED ISSUE #525: SVG.PointArray fail to parse points in Polygon creation
When I create a Polygon I get an error: 
```
svg.js:2537 Error: <polygon> attribute points: Expected number, "221.08,NaN 191.79,NaN 0…".
    attr @ svg.js:2537
    plot @ svg.js:4119
    polygon @ svg.js:4106
```

The easiest way to reproduce is to run the following:
```js
const p = "221.08 191.79 0.46 191.79 0.46 63.92 63.8 0.46 284.46 0.46 284.46 128.37 221.08 191.79"
console.log(new SVG.PointArray(p).toString()) // -> "221.08,NaN 191.79,NaN 0.46,NaN..."
```

To show that this is a valid plot string, you can view the attached `t.svg` which contains a polygon with the exact same points. [t.svg.zip](https://github.com/wout/svg.js/files/506265/t.svg.zip)

The issue seems to be on [line 53 in pointarray.js](https://github.com/wout/svg.js/blob/master/src/pointarray.js#L53):
```js
    // split points
    array = this.split(array)

    // parse points
    for (var i = 0, il = array.length, p, points = []; i < il; i++) {
      p = array[i].split(',') // <- should be "."
      points.push([parseFloat(p[0]), parseFloat(p[1])])
    }
```

For completeness, here is the lines that will trigger the error in Chrome:
```js
const canvas = SVG("canvas")
const p = "221.08 191.79 0.46 191.79 0.46 63.92 63.8 0.46 284.46 0.46 284.46 128.37 221.08 191.79"
canvas.polygon(p)
```

dotnetCarpenter:
This fixes #525 by accepting/parsing coordinate strings where x and y is separated by comma and or by space - we count the shape coordinates until (and including) the maximum even number.

Illustrator exports `points` in the following format: `<polygon class="cls-1" points="221.08 191.79 0.46 191.79 0.46 63.92 63.8 0.46 284.46 0.46 284.46 128.37 221.08 191.79"/>` but SVG.js only support the following format `<polygon class="cls-1" points="221.08,191.79 0.46,191.79 0.46,63.92 63.8,0.46 284.46,0.46 284.46,128.37 221.08,191.79"/>`. Which makes it difficult to implement graphic designers work.

This PR is made upon #527 in anticipation for that being merged, but perhaps I should rebase this against the current master?


dotnetCarpenter:
I rebased against the current master to make this PR more manageable.

dist/svg.js
```diff
@@ -553,16 +553,28 @@ SVG.extend(SVG.PointArray, {
...
+
+    // validate points - https://svgwg.org/svg2-draft/shapes.html#DataTypePoints
+    // Odd number of coordinates is an error. In such cases, drop the last odd coordinate.
+    if (points.length % 2 !== 0) points.pop()
+
+    // parse points as floats
+    points = points.map(function(x) { return parseFloat(x) })
+
+    // wrap points in two-tuples
+    points = points.map(wrapPoints()).filter(function(x) { return x })
```
Fuzzyma: Albeit it looks clean, this code is not very understandable. Its a bit magic going on here.
My idea would be `reduce` or a simple `for loop`. We lately had trouble with performance in svg.js where filter was heavily used.
A loop is still faster than a function call.

dist/svg.js
```diff
@@ -553,16 +553,28 @@ SVG.extend(SVG.PointArray, {
...
-      points.push([parseFloat(p[0]), parseFloat(p[1])])
-    }
+    var points = array.split(/\s|,/)
+
+    // validate points - https://svgwg.org/svg2-draft/shapes.html#DataTypePoints
+    // Odd number of coordinates is an error. In such cases, drop the last odd coordinate.
+    if (points.length % 2 !== 0) points.pop()
+
+    // parse points as floats
+    points = points.map(function(x) { return parseFloat(x) })
```
dotnetCarpenter: You're right - I'll change it

dist/svg.js
```diff
@@ -553,16 +553,28 @@ SVG.extend(SVG.PointArray, {
...
     // parse points
-    for (var i = 0, il = array.length, p, points = []; i < il; i++) {
-      p = array[i].split(',')
-      points.push([parseFloat(p[0]), parseFloat(p[1])])
-    }
+    var points = array.split(/\s|,/)
+
+    // validate points - https://svgwg.org/svg2-draft/shapes.html#DataTypePoints
+    // Odd number of coordinates is an error. In such cases, drop the last odd coordinate.
+    if (points.length % 2 !== 0) points.pop()
```
dotnetCarpenter: According to the [spec](https://svgwg.org/svg2-draft/shapes.html#DataTypePoints):

>If an odd number of coordinates is provided, then the element is in error, with the same user agent behavior as occurs with an incorrectly specified ‘path’ element. In such error cases the user agent will drop the last, odd coordinate and otherwise render the shape.

I think the spirit in the spec is that we should do this silently but we could easily throw an error. SVG.js should be easy to use with drawing tools such as Illustrator and perhaps throwing an error will make it more difficult for users. Someone once said be lenient with input, strict with output.

dist/svg.js
dotnetCarpenter: As long as we don't have more higher order FP functions we can use, I agree that list manipulations quickly grow a bit hairy. I'll change it to a `for loop`.

dist/svg.js
Fuzzyma: Yes sounds right so far. My intention is, that users find mistakes in their js code not in their svg code. Anyway I guess its fine

Fuzzyma:
I really would like to wait for the travis PR to be merged so that we can see it working :). Is travis ready to be merged now?

dotnetCarpenter:
Yes. I believe #527 is ready :-)

dotnetCarpenter:
@Fuzzyma hang on I'll remove `dist/*` from this PR

dotnetCarpenter:
@Fuzzyma Not sure it was a good idea. Well, you need to do `npm run build` and upload the dist  to master after merging this PR

dotnetCarpenter:
@Fuzzyma ok, removing didn't work. I'll rebase with current master

Fuzzyma:
Tests still failing :(

dotnetCarpenter:
Yep, seems we can a slightly different behavior when running Firefox headless. I'll see if I can reproduce it locally, since I don't see these errors when running the unit test manually in the browser.

dotnetCarpenter:
In [spec/element.js I don't see where `draw` is defined](https://github.com/wout/svg.js/blob/master/spec/spec/element.js#L4) thus seems that a lot tests *should* be failing here...

Fuzzyma:
the `sugar.js` tests are not included in the spec runner (why???). Thats why it doenst fail locally... fixing...

Fuzzyma:
the failing tests come from the sugar tests which are made frm me long time ago. I copied from element.js which explains the redundant tests. Also i never finished these tests. Thats why they are failing. I fix this atm

dotnetCarpenter:
@Fuzzyma ok cool. Then I'll do nothing?

Fuzzyma:
Sorry for the delay. I pushed the fix. There were no tests for the sugar module at the start. So I decided to add some but actually forgot about it. This relict was left. Tests are still not complete but valid and passing.

Fuzzyma:
Last test is the `svg()` method which I already fixed for chrome and FF. Looks like linux FF has different attribute ordering to Windows FF. Maybe someone can check this?

Fuzzyma:
Mh - looks like our test bench is not correctly reseted before each test. When running in a Browser the id of the svg is 1001. While it was something different when run in headless browser. We should look into that. However: all tests are passing on my local installation now. Travis should be fine :)

Fuzzyma:
@dotnetCarpenter tests should pass after a rebase. We can merge this then

Fuzzyma:
Do I see a green checkmark there? :D 
Congratulation! You just finished teh first PR which was tested with Travis. Hurray!
Let me now if this can be merged now :)

dotnetCarpenter:
@Fuzzyma YES :rocket: :confetti_ball: !! :dancing_men: 
It's all good. Maybe you want to fix my wordings in the `CHANGELOG.md`. I've tried to write down everything from my last 2 PR's. But everything else looks good to me

Fuzzyma:
Naa - thats something I can do later

pjmolina:
Related to this issue:
I have code such this one: 

```
		draw.polyline("...");   // and
		draw.polygon("...");
````
where the series of points come from Ilutrators with sequences of spaces and commas and separators.

This code worked on version: **2.3.4** 
But now is failing without changes on version: **2.3.5**
				
I got errors of the following type:

```
svg.js?8a91:2541 Error: <polyline> attribute points: Expected number, "…4 2091.6,1353.3 NaN,NaN".
```

Not tested with 3.0.0 yet as 2.3.5 is the last one on npmjs.org
Any fix to be published on npmjs ?
Thxs.

**Comment by Fuzzyma** [comment]
You have code which passes three points to the function?

**Comment by pjmolina** [comment]
Hi Fuzzyma. I rescued real sequeces of points I am passing to the function. Here you have:

`polygon()` sample:  

```
"2176.6,1708.8 2176.4,1755.8 2245.8,1801.5 2297,1787.8  "

	svg.js?8a91:2541 Error: <polygon> attribute points: Expected number, "…8 1802.5,1475.5 NaN,NaN NaN,NaN".
```

`polyline()` sample:

```
"2347,1115 2347,1178.8 2451.5,1164.5 2453.2,1098.2 2348.5,1115.5  "

	svg.js?8a91:2541 Error: <polyline> attribute points: Expected number, "…4.1 2091.6,1536 NaN,NaN".attr @ 
```

Fuzzyma:
Please try the current master and report back. If all is well I will publish it

dotnetCarpenter:
@pjmolina I'm interested too since I broke your code. Does it work now, with [the latest build](https://raw.githubusercontent.com/wout/svg.js/master/dist/svg.min.js)?

pjmolina:
@Fuzzyma @dotnetCarpenter Confirmed! The example provided:

- Works with npm v. 2.3.4
- Fails with npm v. 2.3.5
- Works again at latest commit: [e4bff8c910e37f923a138fd44f6c8a3b21fa2a78](git+https://github.com/wout/svg.js#e4bff8c910e37f923a138fd44f6c8a3b21fa2a78)

You can use the provided point sequences to create a repro unit-test if needed.

Fuzzyma:
We already did that. Actually the big was a stupid mistake on our side. We will try to merge the one pr and then publish a new version

PR #532, opened 2016-10-15
Make SVG.FX.loop modify the last situation instead of the current one

RmiTtro:
I'm not sure if this is a bug, but I think it would make more sense if the loop method in fx.js modified the last situation instead of the current one. This change would make something like this possible:

rect.animate().dmove(150,75).loop(3,true).animate().dmove(50,50).loop(2,true)


Fuzzyma:
mh - intereseting. Yes that could be a bug. The FX Module is very hard to test because of its timing problems. Do you mind adding a test where you show, that the current behavior is wrong? That would be overall useful!

RmiTtro:
As requested, I have added a test that shows that the current behaviour of loop is wrong. Also, since there were no tests at all for the loop method, I took this opportunity to write its full spec :wink:.

Fuzzyma:
You... are... awesome!!
I wonder why the test with "is applied to the last situation" fails. I thought travis tests a specific commit and you actually fixed this case right there. Thats weird...

Fuzzyma:
ah right. Please run `npm install` and `npm build` and commit the changed files

RmiTtro:
Here you go! Sorry about this, I wasn't sure if I should include the new distribution in my commit.

Fuzzyma:
Jeah maybe we should configure travis that it build the lib before running tests...

PR #534, opened 2016-10-16
Fix the test of SVG.FX.afterAll

RmiTtro:
I have found this error while adding tests for the FX module. I also added two call to animate just to make sure that the situations array gets filled.


Fuzzyma:
thx again!

RmiTtro:
You're Welcome!

PR #535, opened 2016-10-17
Fix some bugs on SVG.FX.speed

RmiTtro:
Hi, I have corrected some bugs on the method speed of the FX module:
- change the algorithm to calculate the new duration
- fix speed changing pos when the animation is running backward

These changes should make this method work as it was intended. But before doing anything with this pull request, you should check my other pull request #536 where I offer a different implementation of the
method speed, one that is similar to others animations libraries.


Fuzzyma:
I think we can dismiss this in favour of #536 , okay?

RmiTtro:
Yup, I'm closing it!

PR #536, opened 2016-10-17
Make SVG.FX.speed act more like others animations libraries

RmiTtro:
This pull request makes the speed method of the FX module act more like others animations libraries.

For example, a call .speed(2) would make the animation 2 times faster.
If you then call .speed(4), this would make the animation 4 times faster.
And a call .speed(1) would make it go back to normal speed.

The current implementation is more like this:
A call .speed(2) would make the animation 2 times faster.
If you then call .speed(4), this would make the animation 8 times faster.
And a call .speed(1) would have no effect (the animation stay 8 times faster).

This change would not be backward compatible since it modifies the way the method speed work. Then again, since this method is kind of hidden, it must not be widely used. At the very least, this change could be considered for the next version of this library.

If you decide to refuse this pull request, be sure to check my other pull request #535 that correct some of the bugs with the current implementation of the method speed.

Thx for reading this long comment!


dotnetCarpenter:
I'm new here and have never used SVG.FX but @RmiTtro makes a good argument. His suggested API is more sensible and since we got svg.js 3.0 on the horizon, I think this PR could go into 3.0 and #535 can go into 2.3.5. This is just my opinion and I haven't reviewed the code yet.

spec/spec/fx.js
```diff
@@ -40,12 +41,26 @@ describe('FX', function() {
...
   })
 
   describe('posToTime()', function() {
     it('converts a progress to a timestamp', function() {
       expect(fx.posToTime(0.5)).toBe(fx.situation.start+fx.situation.duration/2)
     })
+
+    it('should take speed into consideration', function() {
+      fx.speed(4)
+      expect(fx.posToTime(0.5)).toBe(fx.situation.start+fx.situation.duration/4/2)
```
dotnetCarpenter: `fx.situation.start+fx.situation.duration/4/2` is not very clear. Can you use variables to better communicate what you're testing here?
The same in the similar test cases. The tests need to be understandable to someone who hasn't written the implementation.

dotnetCarpenter: Everything else looks good at a glance

Fuzzyma:
The horizon is far away ;P. I would prefer to sell this as bugfix :D. I know the behavior changes quite a bit. But:

- FX module is not commonly used
- was modernized a short time ago (speed wasnt available back then)
- does not change the api
- in most cases wont change the behavior because most people use speed once in an animation:

```js
var fx = el.animate().foo().bar() // and never again

// somewhen else
fx.speed(2)
```

So I would suggest addig this to `2.3.6` instead since 3.0 is really far away.
And last but not least:
**The word `speed` is nowheer in the documentation (seems like I forgot to add it). Therefore its not part of the public api and we can change it as we want :P**

__Review of code incoming__

dist/svg.js
```diff
@@ -1541,19 +1542,22 @@ SVG.FX = SVG.invent({
...
     }
 
-    // speeds up the animation by the given factor
-    // this changes the duration of the animation
-  , speed: function(speed){
-      this.situation.duration = this.situation.duration * this.pos + (1-this.pos) * this.situation.duration / speed
-      this.situation.finish = this.situation.start + this.situation.duration
-      return this.at(this.pos)
+    // set/get the speed of the animation
+  , speed: function(spd){
```
Fuzzyma: Are we really sure we want to allow that `speed(0)` will return the speed value? Thats not what I would expect because I passed (an invalid) parameter to the function. I guess we can map `speed(0)` to `stop()` or better `pause()` or just let js throw that division by zero error

spec/spec/fx.js
RmiTtro: You are right, I will make them easier to understand.

spec/spec/fx.js
```diff
@@ -43,10 +43,15 @@ describe('FX', function() {
...
     it('should take speed into consideration', function() {
-      fx.speed(4)
-      expect(fx.timeToPos(fx.situation.start+fx.situation.duration/4/2)).toBe(0.5)
-      fx.speed(0.5)
-      expect(fx.timeToPos(fx.situation.start+fx.situation.duration/0.5/4)).toBe(0.25)
+      var spd
+
+      spd = 4
+      fx.speed(spd)
+      expect(fx.timeToPos(fx.situation.start+(fx.situation.duration/spd)/2)).toBe(0.5)
```
Fuzzyma: writing `* 0.5` instead of `/2` helps to see the relation to `toBe(0.5)`

RmiTtro:
These commits should solve the reported issues. Also, I was wondering if we should add a speed method to SVG.Element so that it can be called directly on the element.

spec/spec/fx.js
```diff
@@ -120,54 +144,83 @@ describe('FX', function() {
...
-      expect(fx.speed()).toBe(0.5)
+    it('pause the animation when a speed of 0 is passed', function(){
+      var spd = fx.speed()
 
-      fx.speed(1)
-      expect(fx.speed()).toBe(1)
+      expect(fx.speed(0)).toBe(fx)
+      expect(fx.speed()).toBe(spd)
+      expect(fx.paused).toBe(true)
     })
```
Fuzzyma: You call 3 times speed in this test while testing for one specific call. That could lead to tests missbehaving.
I suggest just calling `speed(0)` and check if it is paused.

```js
var ret = fx.speed(0)
expect(ret).toBe(fx)
expect(fx.paused).toBe(True)
```

spec/spec/fx.js
```diff
@@ -43,10 +43,15 @@ describe('FX', function() {
...
-      expect(fx.timeToPos(fx.situation.start+fx.situation.duration/0.5/4)).toBe(0.25)
+      var spd
+
+      spd = 4
+      fx.speed(spd)
+      expect(fx.timeToPos(fx.situation.start+(fx.situation.duration/spd)/2)).toBe(0.5)
+
+      spd = 0.5
+      fx.speed(spd)
+      expect(fx.timeToPos(fx.situation.start+(fx.situation.duration/spd)/4)).toBe(0.25)
```
RmiTtro: You are right, I will make the change right now!

src/fx.js
```diff
@@ -337,6 +337,8 @@ SVG.FX = SVG.invent({
 
     // set/get the speed of the animation
   , speed: function(spd){
+      if (spd === 0) return this.pause()
```
Fuzzyma: forgot to mention this: I would rename `spd` to `_speed` followig the convention of `_target`. But that is up to discussion

Fuzzyma:
Looks like one of my comments is lost ins this conversations. Click Files changed to see it

Fuzzyma:
>Also, I was wondering if we should add a speed method to SVG.Element so that it can be called directly on the element

yes, please

RmiTtro:
In these new commits:
- I fix the remaining issues with the tests
- I add speed to SVG.Element
- I add documentation to speed
- I correct a small error in the documentation ;)

Fuzzyma:
There are 2 more comments which are shown as outdated. Did you read them? :)

RmiTtro:
Oops, sorry I had not seen them XD.

spec/spec/fx.js
RmiTtro: The other two calls are just to make sure that the speed hasn't changed. Maybe I should directly check the speed attribute on the fx object?

src/fx.js
RmiTtro: Fine by me, if there is already a convention,  I will follow it! ;)

RmiTtro:
This new commit change the attribute spd to _speed and correct the remaining issues in the tests as pointed out by @Fuzzyma.

dotnetCarpenter:
LGTM

Fuzzyma:
dito

PR #538, opened 2016-10-20
Make SVG.FX.at reverse aware

RmiTtro:
This fix the problem with the method _at_ that when you call _at_ while the
animation is running reversed, the new position is the inverse of the
position passed to _at_. For example, if you do `.at(0.6)` while the animation
is running reversed, the new position would be 0.4 instead of 0.6.

There is also a small change to the method _reset_ that make it reset the
loop counter and the reverse flag.

This should also fix issue #539.


dist/svg.js
```diff
@@ -1540,10 +1542,27 @@ SVG.FX = SVG.invent({
...
+  }
+
     // set the internal animation pointer to the specified position and updates the visualisation
   , at: function(pos){
       var durDivSpd = this.situation.duration/this._speed
 
+      // correct position
+      if(pos > 1) pos = 1
+      if(pos < 0) pos = 0
+      if(this.situation.reversed) pos = 1 - pos
```
Fuzzyma: When I cann `at(0.7)` with `reversed:true`  ecpect the animation to be at the position 0.7 which is the point where it wouldbe when in would run normally without reverse. I then expect the animation to go backwards from that point.
You change the position here. Is my case still working?

dist/svg.js
```diff
@@ -1540,10 +1542,27 @@ SVG.FX = SVG.invent({
...
+    // Cast boolean to number
+    return this.at(+!this.situation.reversed)
+  }
+
     // set the internal animation pointer to the specified position and updates the visualisation
   , at: function(pos){
       var durDivSpd = this.situation.duration/this._speed
 
+      // correct position
+      if(pos > 1) pos = 1
```
Fuzzyma: this constraint is made in the `steP()` method. Why is it needed here?

dist/svg.js
Fuzzyma: I tried to handle all this position juggeling at one place (`step()`). Why is it now needed to do that everywhere?

dist/svg.js
RmiTtro: Well, this constraint and the one just below it are not necessary, they are just a safety measure in case the user of this method tries to pass value out of the range [0,1]. They need to be redone here because of the 3rd _if_ just below that is necessary.

dist/svg.js
RmiTtro: Actually your case was not working in the first place which is why I made this pull request. When you call `at(0.7)` with `reversed:true` it mean that 0.3 of your animation as been done and that is the value you need to use when you modify _start_. Without this line, the position just end up being set to 0.3 and is run backward from there.

Fuzzyma: hm indeed... do you see a way to let the position go backwards instead of 1-pos? That would simplify things a lot. Preferable this is done in the final `step()` methods so other methods do not have to bother about reversing stuff. Just a question. Because I dont have an idea expect calculating the position completely different when animation is reversed :/

RmiTtro: To help you, you can compare it with when the animation is running forward. If I call `at(0.7)` with `reversed:false` it mean that 0.7 of the animation as been done so that is the value that is used to set _start_.

Fuzzyma: jejej I understand that and thats what I want. I just dont like juggling with the position pointer all the time. And I wonder if there is a way to make this cleaner...

RmiTtro: Oups, sorry about the message I sent after your response, it wasn't on Github when I sent it.

Fuzzyma: ohm i might need an example for this. The start code is kinda special and independent from step. Same goes for finish. So how tould you move this to step?

RmiTtro: Actually, I meant the modification of `this.situation.start` and `this.situation.finish`, sorry for the confusion.

Fuzzyma: hm actually this properties are only set twice in the whole code (start and at method). How would moving them help from repetition? :)

RmiTtro: I just want to move them out of `at` so `at` doesn't have to reverse `this.pos`.

Fuzzyma: jeah i know. So what would the step method then looks like?

RmiTtro: Well something like this but hopefully better (I'm on my smartphone so it is hard for me to code XD):

```javascript
 , step: function(ignoreTime){

      // convert current time to position
      if(!ignoreTime) this.pos = this.timeToPos(+new Date)
      else {
        var durDivSpd = this.situation.duration/this._speed

        // correct position
        if(pos > 1) pos = 1
        if(pos < 0) pos = 0
        if(this.situation.reversed) pos = 1 - pos

        this.situation.start = +new Date - pos * durDivSpd
        this.situation.finish = this.situation.start + durDivSpd
      }

      if(this.pos >= 1 && (this.situation.loop === true || (typeof this.situation.loop == 'number' && --this.situation.loop))){

        if(this.situation.reversing){
          this.situation.reversed = !this.situation.reversed
        }

        return this.at(this.situation.reversed ? 1-(this.pos-1) : (this.pos-1))
      }

      if(this.situation.reversed) this.pos = 1 - this.pos

      // correct position
      if(this.pos > 1)this.pos = 1
      if(this.pos < 0)this.pos = 0
```

RmiTtro:
After swirling things around a bit, this the best I managed to achieve in moving the code from SVG.FX.at to SVG.FX.step. Hopefully, this whole change will make the animation library a tiny bit more performant (I have removed the recursion in step). I just hope I didn't hurt readability too much in the process, although I have added some comments to compensate XD.

dist/svg.js
```diff
@@ -1694,20 +1704,53 @@ SVG.FX = SVG.invent({
 
       // convert current time to position
       if(!ignoreTime) this.pos = this.timeToPos(+new Date)
+      else {
+        // If the method at was called with a position above 1, we set it back to 1
+        if(this.pos > 1) this.pos = 1
```
Fuzzyma: Why is this done here? Or better: Why is the other check done outside of the else? I rather would have 2 times the same `if`, when this enables us to group the both checks together

dist/svg.js
```diff
@@ -1694,20 +1704,53 @@ SVG.FX = SVG.invent({
 
       // convert current time to position
       if(!ignoreTime) this.pos = this.timeToPos(+new Date)
+      else {
```
Fuzzyma: while
```js
if (...) {

} else something()
```

is good readable, this one here confuses me a lot.
Better add the brackets in this case

dist/svg.js
```diff
@@ -1694,20 +1704,53 @@ SVG.FX = SVG.invent({
...
+        // If the method at was called with a position above 1, we set it back to 1
+        if(this.pos > 1) this.pos = 1
+        if(this.situation.reversed) this.pos = 1 - this.pos
+      }
 
-      if(this.pos >= 1 && (this.situation.loop === true || (typeof this.situation.loop == 'number' && --this.situation.loop))){
+      if(this.pos < 0) this.pos = 0
+      if(this.pos >= 1){
```
Fuzzyma: didnt we just corrected this case? Well - at least we did for some cases.
Maybe we need to correct it later instead?

dist/svg.js
```diff
@@ -1694,20 +1704,53 @@ SVG.FX = SVG.invent({
...
+          this.situation.loop = currentLoop
+        } else {
+          this.pos = 1
+          ignoreTime = false
         }
-        return this.at(this.pos-1)
       }
 
-      if(this.situation.reversed) this.pos = 1 - this.pos
+      if(ignoreTime){
```
Fuzzyma: did you reuse the function parameter for a different purposes here? Imo that should be avoided. What is this parameter really doing?

dist/svg.js
RmiTtro: I guess you are right, I will add the bracket.

dist/svg.js
RmiTtro: I think I can move `if(this.pos > 1) this.pos = 1` in here, the result would be something like this:
```javascript
if(this.pos >= 1){
  if(ignoreTime) this.pos = 1
```
Maybe that would be better?

dist/svg.js
RmiTtro: I really overused this parameter XD! I have used it here because this part need to also be done when _step_ is called with _ignoreTime_ set to _true_. What I could do to make it more readable is to use a new boolean variable instead, something like _recalculateStart_.

dist/svg.js
Fuzzyma: Why ignoreTime has to be true to cap the value to 1? Imo user can just call `at(1.3)` and it should start the next loop or cap to `1`. Thats why I handled the loop stuff first and then capped the values. I admit that I cannot fully follow your code.
Shouldnt the flow be like this?:

```
calculate position
correct position to reverse if needed
handle overflowing positions (say do the loop stuff)
now cap the values to [0,1]
other stuff...
```

Correct me if i am wrong but for me it looks a bit overcomplicated.

RmiTtro: Well if we allow a call like `at(1.3)` to start the next loop that would definitely simplify thing. But, if we allow that, we would also have to allow, when the animation is reversed, a call like `at(-0.3)` to also start the next loop. Having a negative position looked weird to me, so that is why I limited the values at can receive to between 0 and 1.

dist/svg.js
RmiTtro: Hmm, you are right we could always recalculate and it should work... Really, I can't decide if it is better or not.

dist/svg.js
Fuzzyma: I would rank simple code a lot higher than this issue. Also: allowing more is usually better. The user can decide whether to use it or not

RmiTtro: You convinced me, I will allow _at_ to accept values bellow 0 and above 1.

RmiTtro:
In this commit, I fixed all of the reported issues. I also made start be recalculated every time as suggested and it works, all tests pass and, performance-wise, I don't think it has much of an impact. So, I guess this is a solution we can use even though it seem a little weird to recalculate start every time.

RmiTtro:
I have just thought about this, but recalculating `this.situation.start` every time in _step_ risk to cause problems with the delay (`this.situation.delay`) and the way it is handled in the _start_ method.

Fuzzyma:
Ok, dont risk that. Better save then sorry. I just want clean and easy code. But not by sacrifying the functionality

dist/svg.js
```diff
@@ -1701,12 +1701,12 @@ SVG.FX = SVG.invent({
...
 
       // convert current time to position
-      if(!ignoreTime) this.pos = this.timeToPos(+new Date)
-      else {
-        // If the method at was called with a position above 1, we set it back to 1
-        if(this.pos > 1) this.pos = 1
+      if(!ignoreTime) {
+        this.pos = this.timeToPos(+new Date)
+      } else {
         if(this.situation.reversed) this.pos = 1 - this.pos
```
Fuzzyma: I can spot the same statement at line 1753. It looks like this statement would un-done there

dist/svg.js
```diff
@@ -1740,15 +1738,11 @@ SVG.FX = SVG.invent({
           this.situation.loop = currentLoop
         } else {
           this.pos = 1
-          ignoreTime = false
         }
```
Fuzzyma: I need to take a deeper look into this block. Looks like magic^^
Btw: There is still `if(this.pos < 0) this.pos = 0`. I thought you wanted to allow values below 0

RmiTtro:
I think I have an idea that could make everything a lot simpler, I just need to try it. So don't review this code just yet, hopefully the next commit will be a lot easier to review ;).

Fuzzyma:
Sorry for all this critic. I really appreciate what you are doing here!

RmiTtro:
I think this time it should be good :). I have made a couple of changes and I think it make everything a lot simpler.

dist/svg.js
```diff
@@ -1692,15 +1719,37 @@ SVG.FX = SVG.invent({
...
-        if(this.situation.reversing){
-          this.situation.reversed = !this.situation.reversed
+        if(this.situation.reversing) {
+          // Toggle reversed if an odd number of loops as occured since the last call of step
+          this.situation.reversed = this.situation.reversed != Boolean((this.situation.loop - lastLoop) % 2)
         }
-        return this.at(this.pos-1)
+
+      } else {
+        this.absPos = Math.min(this.absPos, 1)
```
Fuzzyma: this is basically done already a few lines later when doing `if(pos > 1) pos = 1`. Just as notice

Fuzzyma:
Sorry for creating a merge conflict with your other PR. Please rebase. And thanks!!

dist/svg.js
RmiTtro: You are right, I will change it so it is no more performed twice.

RmiTtro:
There you go! All ready to be merged ;)!

Fuzzyma:
Thank you very much for your effort!

RmiTtro:
You're welcome!

PR #540, opened 2016-10-20
replace call to filter function with better regex

dotnetCarpenter:
What do you think @Fuzzyma? I think it's cleaner than f1d9f615


Fuzzyma:
Jetzt. Its a better regex but the problem with leading or trailing spaces remains.
Try:

```
console.log('a b c  '.split('/s+')
```

You should see it then

dotnetCarpenter:
Well, for that we could easily do `console.log(' a b c  '.trim().split('/s+'))`. But then the question is if we prefer
```js
// parse points
array = array.trim().split(/\s+|,/)
```
over
```js
// parse points
array = array.split(/\s|,/).filter(function(a){ return a !== '' })
```

Fuzzyma:
trim is a good solution. Go with yours + trim

dotnetCarpenter:
Hmm... something weird is going on with my commits and github. The current commit on travis is not my latest commit...

dotnetCarpenter:
ahh, it was a conflict. Fixed

PR #543, opened 2016-11-01
Make matrixify work with transformation chain separated by commas

RmiTtro:
According to the SVG spec, transformation chain can be separated by whitespace
and/or commas. The method matrixify was not working with transformation chain
separated by commas. This commit should fix that bug.


Fuzzyma:
thx

RmiTtro:
You're welcome

PR #545, opened 2016-11-02
Fix the implementation of the skew transform

RmiTtro:
While trying to fix another bug, I spotted a couple of issues with the current implementation of the skew transform. This commit should fix all of them. It also fix a bug where when calling scale with 3 parameters, cx is not set to the right value.


dist/svg.js
```diff
@@ -2414,15 +2415,28 @@ SVG.Matrix = SVG.invent({
...
+        cy = cx
+        cx = y
+        y = x
+      }
+
+      // convert degrees to radians
+      x = SVG.utils.radians(x)
+      y = SVG.utils.radians(y)
+
+      return this.around(cx, cy, new SVG.Matrix(1, Math.tan(y), Math.tan(x), 1, 0, 0))
```
Fuzzyma: why is in your eyes the native `skew` function not working correctly?

RmiTtro: When you chain skewX with skewY you end up also modifying the _a_ value  of the matrix which is not desirable (took me some time to figure out this issue XD). Unfortunately, there are no native skew that performs a skewX and skewY simultaneously, so that is why I ended up defining it this way.

dist/svg.js
```diff
@@ -2414,15 +2415,28 @@ SVG.Matrix = SVG.invent({
...
     }
     // SkewX
   , skewX: function(x, cx, cy) {
-      return this.around(cx, cy, this.native().skewX(x || 0))
```
RmiTtro: There was also an issue with line, it should have been something like this (was REALLY HARD to spot this one XD):
```javascript
return this.around(cx, cy, new SVG.Matrix().native().skewX(x || 0))
```
Since no other transform (scale, rotate, ...) use the native function, to fix it, I just made it call skew.

dist/svg.js
```diff
@@ -2414,15 +2415,28 @@ SVG.Matrix = SVG.invent({
...
+      return this.skew(x, 0, cx, cy)
     }
     // SkewY
   , skewY: function(y, cx, cy) {
-      return this.around(cx, cy, this.native().skewY(y || 0))
```
RmiTtro: Same thing for this one

dist/svg.js
Fuzzyma: And you are sure, that this will work for all valid values of skew?

RmiTtro: Well it should, I did pick this matrix from the spec:
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform
https://www.w3.org/TR/SVG/coords.html#EstablishingANewUserSpace

RmiTtro: Something I should really have mentioned is the problem that this is fixing. The following code should illustrate:

```javascript
var rect = draw.rect(100,100)

// Before this commit
rect.skew(14, 20)
rect.transform().skewX // return 14 as expected
rect.transform().skewY // return 18.453269962755858 instead of 20

// With this commit
rect.skew(14, 20)
rect.transform().skewX // return 14 as expected
rect.transform().skewY // return 20 as expected
```

Fuzzyma:
Well the values from the transform function are rather unreliable because
you can't say from the matrix which plain values where set before. You can
only guess. But in this case it seems legit :)

Am 03.11.2016 02:05 schrieb "Rémi Tétreault" <notifications@github.com>:

RmiTtro:
What made me search for a fix was that skewX was right on while skewY was just close. I was finding that weird, I mean either both are right on or both are just close, but not one being right while the other is not.

Fuzzyma:
I guess when you would switch the both transformations, the values vor the
right and wrong one would change too. Still silly. I guess this one is way
better!

Am 03.11.2016 23:11 schrieb "Rémi Tétreault" <notifications@github.com>:

PR #546, opened 2016-11-03
Fix and improve the dequeue method of the FX module

RmiTtro:
The first improvement is to stop using setTimeout to perform the delay. It is
now performed in a manner similar to the start method. This should be a lot
more reliable than using setTimeout and allow some nice goodies like
being able to pause it or affect its duration using the speed method.

Also, fix a bug where when dequeuing a function, the function was executed, but
the next elements of the queue where not handled.

The second improvement is to have the delay method add an empty situation
(with its duration set to the duration of the delay) to the queue. This change
allows dequeue not to have to treat delay added to the queue as something
special, now its just a situation.


Fuzzyma:
Ich hab the same thoughts about the delay when I rewrote the fx module. The only reason I didn't do it was because it was that way before. I fully agree on this idea. Review coming soon!

dist/svg.js
```diff
@@ -1409,23 +1415,14 @@ SVG.FX = SVG.invent({
...
-          setTimeout(function(){fn()}, this.situation.delay)
-        }else{
-          fn()
+        if(this.situation instanceof SVG.Situation) {
+          this.startCurrent()
+        } else {
+          // If it is not a SVG.Situation, then it is a function, we execute it
+          // and then pull the next element from the queue
+          this.situation.call(this)
+          return this.dequeue()
```
Fuzzyma: there is a good reason not to have `dequeue` called here. The reason is simply, that `dequeue` should be called **inside** the function. So this function can do asyncronous operations (like getting data via ajax or just animating some other stuff) and call `dequeue` when its done.
So this was intended and should be done imo since you loose any control as a user when dequeue is called immeditaley

RmiTtro: I see your point, I was seeing this as a way to add a callback after an animation but come to think of it there is already `after` that does this... I will change it back :flushed:

RmiTtro:
SVG.Delay is already out of the picture, as for the timeout parameter, it wasn't defined in the first place.

RmiTtro:
I removed the call to `dequeue` after executing a function.

Fuzzyma:
I wonder why it always shows only one commit when you are doing a PR. Whats with the changes you did after the initial PR commit? They are getting just merged into the old one. Any idea why?

RmiTtro:
Well, that is actually my fault XD, I amended my last commit. I did this to have a clean history. So instead of having a commit that make a change followed by an another commit that fix some stuffs on the change you just have the commit with the change as if it was written right the first time it was submitted. If you don't like me doing this just say so and I will stop.

Fuzzyma:
Mh I guess for this one it is fine. I just was a bit confused :D

Am 05.11.2016 01:39 schrieb "Rémi Tétreault" <notifications@github.com>:

PR #548, opened 2016-11-08
Fix a bug with the animation of a scale transform

RmiTtro:
Fix the bug where animating a scale transform also cause a translation when there is already a transform in place. This bug is illustrated here: https://jsfiddle.net/ockc9u5x/1/.


Fuzzyma:
In the jsfiddle the box is scaled around its origin which lay in 0,0.
Standard behavior from svg.js is origin in the center of the shape right?
Because if not, the stated behavior is actually right :D.
Nevertheless I see your point!

Fuzzyma:
I see an issue here. See the following chain of transformations:

```js
el.scale(2, 0, 0).scale(0.5, 50, 50)
```

When `scale(0.5)` is called, it has to undo the scale of the element since it is an absolute transformation.
So it will take the current transformation (which is ``scale(2)` around `0,0`) and will undo it. But it will undo it with its own origin. So it does an `undo` scale of factor `2` around `50,50` instead of `0,0`.
Is this what the user expects?

Absolute transformations are crap (ever been). Matrices dont work like that. However duo to backwards compatibilty (and because of groups which do not have x and y) we have it in svg.js.
When setting an absolute transformation the user has to be aware that there can be "mistakes" (actually you cant really call it mistakes because there are many valid solutions to one problem here). We can only try to make it closer to what the user expects.

Long story short: Does this PR improves the behavior which the user expect in this case (and in different similar cases)? Can we construct an example where this solution actually does more harm then good?

RmiTtro:
Glad I'm not the only one who think absolute transform are crap XD. As for this PR, I may have a better way to solve this problem, I just need to test it out.

RmiTtro:
Forget what I said in my previous message, what I tried to do didn't make a difference XD. The commit I pushed just fix other issues I found while trying my solution. I still think this PR should be merged in since it make animated absolute transform perform the same way than absolute transform applied directly on the element. I tried to find a way to make absolute transform less buggy, but unfortunately I didn't come up with anything :(.

Fuzzyma:
I think you have a point there with the consistency

PR #549, opened 2016-11-11
Fix going down prototype chain

Toosick:
This fixed the a.undo is not a function error. The for loop was going down into  the array's prototype chain.

### DIFF
--- a/src/fx.js
+++ b/src/fx.js
@@ -633,7 +633,7 @@ SVG.FX = SVG.invent({
 
         // get inital initialTransformation
         at = s.initialTransformation
-        for(i in s.transforms){
+        for(i = 0 ; i < s.transforms.length ; i++){
 
           // get next transformation in chain
           var a = s.transforms[i]


dotnetCarpenter:
@Toosick Can you post some code that broke fx.js before your fix or even better write a small unit test in `spec/fx.js`?

dotnetCarpenter:
@Toosick Forgot to say thanks for your bug fix. I'm a little new to fx.js and are still trying to reproduce the error you're fixing. `eachAt` is undocumented so I'm guessing you're seeing this when `eachAt` is called from `step`. It would be interesting get a gist of your failing animation.

Fuzzyma:
The Problem is, that we iterate an array with `in` which is very bad practice (sorry for this).
If you then extend the Array prototype, its also iterated in the loop and whatever this is, it has no method called `undo`.

Example for clarification:

```js

// extendting the array Prototype
// which is btw also bad practice
Array.prototype.printLength = function(){
    console.log(this.length)
}

a = [1,2,3]
for(var i in a)
    console.log(a[i])

// prints 1,2,3,function
```

dotnetCarpenter:
Yes, I understand this. I was just trying to see if I could add a test case for this before merging. I guess my real question is, if this is hypothetical or a practical issue?
But in any case we should include the fix.

Fuzzyma:
There are plenty of libs who sadly still extend native prototypes. So it can get practical very fast.

But I dont even know how to test on this...

dotnetCarpenter:
@Fuzzyma I'll do it. Got some time anyway

dotnetCarpenter:
@Fuzzyma I had to revert my optimization. I forgot to build before I tested and just realized that it was failing.

After closer inspection, it seems that we should leave most (if not all) of the `i in` loops as is, because we are iteration over object, not array.

I could add
```js
// skip properties defined in the prototype chain
if(!s.animations.hasOwnProperty(i)) continue
```
but it seems like a pointless check. unless we think people would actually add to `Object.prototype`.
In that case I think we are better off with a iterator function we can re-use across all of SVG.js - this function would be a hot loop and get JIT optimized in all current browsers.

dotnetCarpenter:
To prevent myself from pushing failing builds again I have create a [pre-push hook](https://gist.github.com/dotnetCarpenter/bb6c2f040e11922cc8f81ef42a8054fd). Maybe it's useful for others than me. I'll add it to the [wiki](https://github.com/svgdotjs/svg.js/wiki/).

Fuzzyma:
Autsch! I should really think about what I am writing. I already wondered
why i used all this loops...
But isn't the loop with transformations an array?

Am 12.11.2016 23:01 schrieb "Jon Ege Ronnenberg" <notifications@github.com>:

dotnetCarpenter:
> But isn't the loop with transformations an array?
Yes! That one is.

Toosick:
@dotnetCarpenter I don't know if you still need my response. I am using SVG.js in conjunction with EmberJS (http://emberjs.com/) and ES6. 

EmberJS extends the array prototype with a ton of methods.

I don't usually use for loops like I did in this fix (really only use the for(let item of items) ES6 syntax now), but since this project isn't using ES6 I made it compatible with what you guys are using.

dotnetCarpenter:
@Toosick thanks for the info. I have merged you commit in (3176d70). It is nice to how a concrete idea about why fixes are needed - both to not repeat past mistakes but also in order to construct automated tests that ensure SVG.js works well in the wild.

Mutating the `Array.prototype` is definitely something we could do in a stress test.

I'm afraid we are a bit far from using ES6 features. Personally I would like to use `import`/`export` with a new bundler instead of just blindly concatenating all our source files. This would also mean we would be able to:

1. only test changes, instead of everything
1. show test status in a IDE while developing
2. easily integrate a precompiler so we can use ES2015-16-17 etc. features
3. build SVG.js with only the parts you use in a project and eliminate all dead code in relation to your project (tree shaking)
4. enable static analysis of our code base, to do all kind of fun tricks

If you can, I would like to see what you built with Ember and SVG.js.

Toosick:
Well I didn't use the Ember aspect as much in this project as it is pretty basic. https://www.youtube.com/watch?v=wc6Dwhb6Ye0

Issue #139, opened 2014-02-03
Does not seem to work in a stand-alone SVG


vadi2:
Having included svg.js in a stand-alone SVG and added the basic draw example code to be run off the DOMContentLoaded event, I'm getting an error in the following code:

```
  /* If the target is an svg element, use that element as the main wrapper.
       This allows svg.js to work with svg documents as well. */
    this.constructor
      .call(this, this.parent.nodeName == 'svg' ? this.parent : SVG.create('svg'))
```

`Uncaught TypeError: Cannot read property 'nodeName' of null`

This is using Google Chrome 32.0.1700.102 on Ubuntu and the latest version of svg.js downloaded today.

Spaxe:
+1. I bumped into this issue right away. This should be in the getting started page.

imanzuk:
+1 for this: "I bumped into this issue right away. This should be in the getting started page."

cassidoo:
+1, just bumped into this issue myself.

Fuzzyma:
I dont want to say its written in the first chapter of the readme but... its written in the first chapter of the readme!

```
The first argument can either be an id of the element or the selected element itself.
```

sverrirs:
Yeah so I had this problem as well today and I read the docs. 

This issue is absolutely **NOT** clear in the docs and NONE of the example code show this. They all indicate that the necessary DOM structure will be created for you, i.e:

> Create an SVG document
> Use the SVG() function to create an SVG document within a given html element:
> var draw = SVG('drawing').size(300, 300)
> var rect = draw.rect(100, 100).attr({ fill: '#f06' })
> The first argument can either be an id of the element or the selected element itself.
> This will generate the following output:
>``` <div id="drawing">
>     <svg xmlns="http://www.w3.org/2000/svg" version="1.1" xmlns:xlink="http://www.w3.org/1999/xlink" width="300" height="300">
>         <rect width="100" height="100" fill="#f06"></rect>
>     </svg>
> </div> ```

This is a simple fix to the docs and users will not get frustrated and give up on this project.

Fuzzyma:
I am sorry to hear that.

> This is a simple fix
What exactly?

sverrirs:
Adding a section such as the one below near the top of the docs and mention something along the lines of "given this HTML" and then go on discussing the API.

```
<svg id="drawing">
</svg> 
```

dotnetCarpenter:
[...]

PR #561, opened 2016-12-20
Implement the morph method of SVG.PathArray

RmiTtro:
This pull request adds an implementation to the method morph of SVG.PathArray. When calling morph, if the passed path uses the same commands in the same order as the object's path, then it is easy to animate between the two as mentioned in the [SVG spec](https://www.w3.org/TR/SVG11/paths.html#PathElement). 

But, when the two paths differ, it is a bit more complex as a conversion need to be performed on the two paths for them to have the same commands. In order to implement this conversion operation, I have used the plugin Interpolate of Inkscape as a reference. Basically, the approach of this plugin is to convert all the segments of the two paths into cubic Bezier curves and then add points on them as necessary for them to have the same commands. 

The implementation of this algorithm is in its own file (_makepathsmorphable.js_) as it is quite lengthy. And to use it, I have added its function (_makePathsMorphable_) in the SVG.utils object.

To see the it in action: http://plnkr.co/edit/9gjlHvtYbisruPxEZxAu?p=preview.


Fuzzyma:
Did you took a look at https://github.com/Fuzzyma/svg.pathmorphing.js? This addon seems to achieve what you want while also allowing different lengths of path commands. Well - but it cannot convert arcs.

Yours looks more clean and has same/more/less features :D. I am biased - dunno what to do ;)

RmiTtro:
Damn, I didn't know you had already one implemented XD. I will have to take
a look at it!

wout:
We chose to provide this as a plugin to keep the size of the library down.

RmiTtro:
It makes sense, I must admit that it requires a lot of code to make this work. But may I suggest that in the library we had some basic functionality for when the paths use the exact same commands. This just require 2 small for loop (basically what I made in the method _haveSameCommands_ and _at_ of the class SVG.PathArray). This basic functionality is the way to go in the [SVG spec](https://www.w3.org/TR/SVG11/paths.html#PathElement) when animating path.

RmiTtro:
In this last commit, I remove almost all of my changes and I just keep the code that was in the method _at_ and the method _haveSameCommands_. As for _morph_, I implemented the most basic algorithm possible: set the _destination_ attribute only if the passed path uses the same commands in the same order as the object's path. I also keep 2 changes that I made in the SVG.Point class. The first one allows setting both x and y by passing only one number to the SVG.Point constructor. The second one allows passing x and y coordinates to the _morph_ method.

As for the rest of my code, I think I will move it into a plugin so that all my work don't go into the trash XD.

Fuzzyma:
@wout @dotnetCarpenter - So what do you guys think? Imo its a straight forward implementation which does not take more space then other morph methods. Also its well tested. Plugins always can overide behavior so when someone needs another implementation he can just use a plugin

wout:
My only concern is size and since it's not too heavy, I suggest we go ahead and include it into the core.

dotnetCarpenter:
I have only skimmed the code but there is a couple of places that perplex me. I don't really have time to go trough it until Friday, though.

src/patharray.js
```diff
@@ -100,6 +100,63 @@ SVG.extend(SVG.PathArray, {
 
     return this
   }
+  // Test if the passed path array use the same commands as this path array
+, haveSameCommands: function(pathArray) {
```
dotnetCarpenter: Not sure why this is needed. Are we looking for a base class, if so then we could use `instanceof` or do we want to duck type `pathArray`?

src/point.js
```diff
@@ -23,9 +23,9 @@ SVG.Point = SVG.invent({
       return new SVG.Point(this)
     }
     // Morph one point into another
-  , morph: function(point) {
+  , morph: function(x, y) {
       // store new destination
-      this.destination = new SVG.Point(point)
+      this.destination = new SVG.Point(x, y)
```
dotnetCarpenter: Changing signature but I don't see the signature is changes in the `SVG.Point` constructor.

dotnetCarpenter:
@RmiTtro I'm not against this in any way. Just need some time to understand it before I say ok. But if @wout and @Fuzzyma say it's fine then I trust them, that it's fine.

src/point.js
RmiTtro: Well this change is just to allow _morph_ to be passed the coordinates as parameters. With this change, all of the following are valid ways to call _morph_: 

```javascript
pt.morph({x:3, y:4}) 
pt.morph([3, 4])
pt.morph(3, 4)
```
There is no need to make any changes to the constructor since it can already be passed two parameters. In fact, I made this change mostly to make the method _morph_ more consistent with the constructor.

src/patharray.js
RmiTtro: Don't worry, it's nothing that messy! This method just compares 2 path array to see if they use the same path data commands (commands like: M, C, A, etc...). Looking back at the name of this method, I can see how it can be misleading. Maybe I should change it to something else like: _useSameCommands_, _useSamePathCommands_, _useSamePathDataCommands_ or something else, I'm open to suggestion.

wout:
I'll have to take a closer look at the functionality. If you can only morph between paths with equal point lengths, it's not a very useful implementation. I'll let you know something later today.

Fuzzyma:
Its the most simple and most fool prooven way to morph paths. As already mentioned: Thats the way svg does path morphing which is also stated in the specs. I see it as fast and simple. Whoever needs more just can use a plugin instead

wout:
Ok, I suppose that's a welcome addition then. Just checked, SVG.js is `16K` gzipped, Snap is `27K`. So we have space left to add some goodness. :)

dist/svg.js
```diff
@@ -2499,9 +2557,9 @@ SVG.Point = SVG.invent({
       return new SVG.Point(this)
     }
     // Morph one point into another
-  , morph: function(point) {
+  , morph: function(x, y) {
```
Fuzzyma: Allows a nice shortcut so I guess its fine ^^

dist/svg.js
```diff
@@ -705,6 +705,63 @@ SVG.extend(SVG.PathArray, {
...
+    return haveSameCommands
+  }
+  // Make path array morphable
+, morph: function(pathArray) {
+    pathArray = new SVG.PathArray(pathArray)
+
+    if(this.haveSameCommands(pathArray)) {
+      this.destination = pathArray
+    } else {
+      this.destination = undefined
```
Fuzzyma: is this needed? I mean - its undefined anyway. And sice every dude can do something like `undefined = true`, its rather uncommon to use `undefined` in an assignment.

dist/svg.js
```diff
@@ -705,6 +705,63 @@ SVG.extend(SVG.PathArray, {
 
     return this
   }
+  // Test if the passed path array use the same commands as this path array
+, haveSameCommands: function(pathArray) {
```
Fuzzyma: In case you wanna rename it: Something like `equalCommands`?
If not: I, at least, would name it `hasSameCommands`. Because its one patharray which `has` the same commands

dist/svg.js
RmiTtro: This is for the case when a user call `morph` multiple times, it is to prevent the attribute `destination` to remain set to a previous value when the test `haveSameCommands` is unsuccessful.  Maybe I should use `null` instead?

dist/svg.js
RmiTtro: `equalCommands` sounds good to me!

dist/svg.js
Fuzzyma: jep - `null` is imo better

dist/svg.js
RmiTtro: Glad you like it!

RmiTtro:
The new commit rename the method haveSameCommands to equalCommands and
replace undefined by null.

dotnetCarpenter:
@RmiTtro I'll try to look this through tonight.

dotnetCarpenter:
Just wondering if we should throw a meaningful error message if the commands are not the same. Maybe hint about https://github.com/Fuzzyma/svg.pathmorphing.js

Simple things going wrong has been [confusing for people](https://github.com/svgdotjs/svg.js/issues/139#issuecomment-63479395) before and if I just copy/paste from our [documentation](http://documentup.com/svgdotjs/svg.js#path) and remove one parameter for a command, I get lots of `Error: <path> attribute d: Expected number, "…100 900 102 898 NaN ".`

Come to think of it, shouldn't `equalCommands` be a guard against such errors, @RmiTtro?

dotnetCarpenter:
Other than that it looks good to me.

RmiTtro:
I don't think its `equalCommands` role to be a guard against those errors as they happen way before `equalCommands` is called. Also, don't forget that `equalCommands` work on paths that have passed through the SVG.PathArray constructor, so if we want to perform some verification on the path, that would be the place to do it.

I would also like to add that it is to be expected that passing a bad formatted path string return an error, especially for the code you copied from the doc since it try to render the passed path string. But, should svg.js implement its own set of custom exceptions is another question entirely. On the positive side, it would allow svg.js to return more meaningful error messages that would help users of this library in their debugging process. On the negative side, that would be a lot of work to implement. That might be something to consider, but that would definitely require its own pull request.

dotnetCarpenter:
@RmiTtro You're right. Now I've played around with the code a little and stepped through the debugger, I see what's going on.

I like to write code a little more terse but that might be unreadable to others. E.g.
Instead of writing this:
```js
morph: function(pathArray) {
  pathArray = new SVG.PathArray(pathArray)

  if(this.equalCommands(pathArray)) {
    this.destination = pathArray
  } else {
    this.destination = null
  }

  return this
}
```
I would write:
```js
morph: function(pathArray) {
  pathArray = new SVG.PathArray(pathArray)

  this.destination = this.equalCommands(pathArray) ? pathArray : null

  return this
}
```

I think we should merge in this PR. Who wants to do the honors? :santa:
Happy holidays to your all :deer: 

Oh, I almost forgot. We should really document this. Both because it's a new feature but also because we have a plugin (maybe soon two - @RmiTtro?) that does much more work in case the user gets put off by our mini implementation.

wout:
@dotnetCarpenter You go ahead sir :)

As for documentation, I'm nearly ready with the new SVG.js website which really is a documentation page. More on that soon!

Merry everything and Happy always!

Fuzzyma:
> I like to write code a little more terse but that might be unreadable to others.

I love doing that but only when it remains clean. In case of the :? operator I personally prefer the good old if-else. But whatever: merge this already and happy xmas! :P

RmiTtro:
Personally, I also like to use ?: operator. I didn't use it in this case because, originally, that `if` had more code in it and when modifying my changes I decided to let it live on in case I would have to add stuff to it again.

Anyway, thanks for accepting my pull request and the rest of my code is definitely going into a plugin sometime in the future.

Merry xmas to you all! :christmas_tree:

Issue #564, opened 2017-01-06
Missing signature/property in TS typing


albohlabs:
```SVG('drawing')``` function call produces the following error:

```
error TS2349: Cannot invoke an expression whose type lacks a call signature. Type 'typeof "svg.js"' has no compatible call signatures.
```

The ```SVG.Matrix``` access results into another error message:

```
error TS2339: Property 'Matrix' does not exist on type 'typeof "svg.js"'.
```

albohlabs:
Import the lib via ```import * as SVG from 'svg.js'```.

@rdfriedl Do you have any idea what i'm doing wrong?

Fuzzyma:
The most likely thing is that you are doing everything right but the signature is wrong. We are missing the knowhow to keep this up to date. Feel free to file a pull request to fix those issues!

rdfriedl:
If the typescript modules work anything like ES6 modules, then i have a good idea of whats going on.

If you do `import * as <some name> from "<some lib>"` then it takes all the exports from the library and dumps them into an object.
Though since svg.js is not written in ES6 or typescript there is only one export, the `SVG()` function.
So when you do `import * as SVG from "svg.js"` you get something that looks like this
`{ SVG: SVG() // main SVG function }`

Again if the typescript modules work the same as ES6 then the correct way to import it would be `import SVG from "svg.js"`

Hope that helps
`

albohlabs:
Thanks for you help.
I adapt the export declaration from [lodash](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/lodash/index.d.ts#L239), [angular](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/types-2.0/angular/index.d.ts#L17) and the [module template](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html) and this works fine with both imports.

See PR #565 for more infos.

PR #565, opened 2017-01-06
Add UMD module support Add Matrix property

albohlabs:
- Edit export as recommended in typescript [templates](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html) and [lodash](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/lodash/index.d.ts#L239) and [angular](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/types-2.0/angular/index.d.ts#L17) typings.
- Add `Matrix` property

Fix for #564

### DIFF
--- a/svg.js.d.ts
+++ b/svg.js.d.ts
@@ -1,3 +1,8 @@
+export = svgjs;
+export as namespace svgjs;
+
+declare var svgjs: svgjs.Library;
+
 // todo add SVG.FX
 declare namespace svgjs {
     export interface Library {
@@ -886,9 +891,4 @@ declare namespace svgjs {
 
         // TODO style, etc, bbox...
     }
-}
-
-declare var SVG: svgjs.Library;
-declare module "svg.js" {
-    export default SVG;
 }
--- a/svg.js.d.ts
+++ b/svg.js.d.ts
@@ -77,7 +77,7 @@ declare namespace svgjs {
 
     // boxes.js
     export interface BBox {
-        new (element?: Element)
+        new (element?: Element): BBox;
         height: number;
         width: number;
         y: number;
@@ -86,8 +86,12 @@ declare namespace svgjs {
         cy: number;
         merge(bbox: BBox): BBox;
     }
-    export interface RBox extends BBox { }
-    export interface TBox extends BBox { }
+    export interface RBox extends BBox {
+        new (element?: Element): RBox;
+    }
+    export interface TBox extends BBox {
+        new (element?: Element): TBox;
+    }
     interface Container {
         bbox(): BBox;
         rbox(): RBox;
@@ -100,7 +104,9 @@ declare namespace svgjs {
     }
 
     // clip.js
-    export interface ClipPath extends Container { }
+    export interface ClipPath extends Container {
+        new (): ClipPath;
+    }
     interface Container {
         clip(): ClipPath;
     }
@@ -113,6 +119,7 @@ declare namespace svgjs {
 
     // color.js
     export interface Color {
+        new (): Color;
         new (color: string): Color;
         new (color: Color): Color;
         r: number;
@@ -131,6 +138,7 @@ declare namespace svgjs {
 
     // container.js
     export interface Container extends Parent {
+        new (): Container;
         viewbox(): ViewBox;
         viewbox(v): this;
     }
@@ -174,11 +182,14 @@ declare namespace svgjs {
     }
 
     // defs.js
-    export interface Defs extends Container { }
+    export interface Defs extends Container {
+        new (): Defs;
+    }
     interface Library { Defs: Defs }
 
     // doc.js
     export interface Doc extends Container {
+        new (): Doc;
         new (selector: string): Doc;
         new (domElement: HTMLElement): Doc;
         namespace(): this;
@@ -187,12 +198,11 @@ declare namespace svgjs {
         spof(spof): this;
         remove(): this;
     }
-    interface Library {
-        Doc: Dec;
-    }
+    interface Library { Doc: Doc; }
 
     // element.js
     export interface Element {
+        new (): Element;
         node: LinkedHTMLElement;
         type: string;
 
@@ -269,8 +279,12 @@ declare namespace svgjs {
         size(width: any, height: any): this;
         radius(x: number, y?: number): this;
     }
-    export interface Circle extends CircleMethods { }
-    export interface Ellipse extends CircleMethods { }
+    export interface Circle extends CircleMethods {
+        new (): Circle;
+    }
+    export interface Ellipse extends CircleMethods {
+        new (): Ellipse;
+    }
     interface Container {
         circle(size?: any): Circle;
         ellipse(width?: any, height?: any): Ellipse;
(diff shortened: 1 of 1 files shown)


Fuzzyma:
@rdfriedl can you review this please? :)

svg.js.d.ts
```diff
@@ -1,3 +1,8 @@
+export = svgjs;
+export as namespace svgjs;
+
+declare var svgjs: svgjs.Library;
+
```
rdfriedl: could you explain what this code dose?
because when i test the typings with this code it just says that the library dose not have a default export.

svg.js.d.ts
```diff
@@ -429,8 +434,14 @@ declare namespace svgjs {
...
         matrix: Matrix;
     }
+
+    interface Library {
+        Matrix: Matrix;
+    }
+
     export interface Matrix {
         (source: any): Matrix;
+        new(source: any): Matrix;
```
rdfriedl: this makes sense though i think it highlights an even bigger problem with the typings.

when i wrote the typings i forgot to supply the arguments to the methods on the library
what i mean is, this works
```typescript
var matrix: svgjs.Matrix = new SVG.Matrix();
```
but this dose not
```typescript
var element: svgjs.Element = new SVG.Element();
var matrix: svgjs.Matrix = new SVG.Matrix(element);
```
this is a mistake on my end since i set the `Matrix` as a function on the `Library` and not a property that holds a reference to the `Matrix` class
a few examples, [here](https://github.com/svgdotjs/svg.js/blob/4770c1e3b817f0ded7403f35edb4f724bf36af2c/svg.js.d.ts#L461) and [here](https://github.com/svgdotjs/svg.js/blob/4770c1e3b817f0ded7403f35edb4f724bf36af2c/svg.js.d.ts#L366)
but i remembered to include in other places, like [here](https://github.com/svgdotjs/svg.js/blob/4770c1e3b817f0ded7403f35edb4f724bf36af2c/svg.js.d.ts#L366)

of course this means that we would have to run over the whole library and fix the parts like this

```typescript
interface Library { Number(): void; }
```
and change then to something like this
```typescript
interface Library { Number: Number; }
```
and then fix this
```typescript
export interface Number{
    // change this
    (value: any, unit?: any): Number;
    // to this, that way SVG.Number can only be called with the "new"
    new (value: any, unit?: any): Number;
}
```
so while this dose fix it for the Matrix class i think we will end up having to go over the entire file and fix each class

rdfriedl:
while it dose not completely fix the bug, it dose highlights a big problem with they typings.
and that is
```typescript
// while this works
var matrix: svgjs.Matrix = new SVG.Matrix();
// this dose not
var matrix: svgjs.Matrix = new SVG.Matrix(1,0,0,1,0,0);
```

i think the typings are going to need a little bit more work to fix and update them before they match.

also its partly my fault that this bug is here, when i was reviewing the code i found two typos [here](https://github.com/svgdotjs/svg.js/blob/4770c1e3b817f0ded7403f35edb4f724bf36af2c/svg.js.d.ts#L713) and [here](https://github.com/svgdotjs/svg.js/blob/4770c1e3b817f0ded7403f35edb4f724bf36af2c/svg.js.d.ts#L713)

svg.js.d.ts
albohlabs: The reason for that is that `svg.js` does not contain a `default` export. As a result we need to import the entire module content as an object, which is the corresponding es6 syntax. [1]

> could you explain what this code dose?

> … anywhere the module keyword was used when declaring an internal module, the namespace keyword can and should be used instead. [2]

It makes the `svgjs` namespace available globally if you use the `/// <reference` directive. [3]

The rest of the code is doing the same as before, if you replace the word namespace with module.

[1] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#Description
[2] https://www.typescriptlang.org/docs/handbook/namespaces.html
[3] https://github.com/Microsoft/TypeScript/issues/7125

rdfriedl: i did some testing with importing the library and it looks like
```typescript
import SVG from 'svg.js'
```
and
```typescript
import * as SVG from 'svg.js'
```
returns the default `SVG()` function.
Its also possible to do 
```typescript
import {Element} from 'svg.js'
```
with that in mind I tested the old typings file and the one here and this is what i got.

**The original**
this worked fine
```typescript
import SVG from 'svg.js'
var draw: svgjs.Doc = new SVG.Doc();
```
this dose not, because the original typings only exports `svgjs.Library`
```typescript
import {Doc} from 'svg.js'
var draw: svgjs.Doc = new Doc();
```
this dose not work.
it makes SVG a object with the `default` property set to the default `SVG()` function
```typescript
import * as SVG from 'svg.js'
var draw: svgjs.Doc = new SVG.Doc();
```

**With your version**
this dose not work, it says that there is no default export
```typescript
import SVG from 'svg.js'
var draw: svgjs.Doc = new SVG.Doc();
```
but this works fine.
except for the `svgjs.Doc`, it says that `Namespace 'svgjs' has no exported member 'Doc'`
```typescript
import {Doc} from 'svg.js'
var draw: svgjs.Doc = new Doc();
```
also with your version this works.
but has the same problem with the svgjs namespace
```typescript
import * as SVG from 'svg.js'
var draw: svgjs.Doc = new Doc();
```

**Conclusion**
it looks like your version works a little better then the original.
we would just have to fix the default export and the namespace error.

albohlabs: Great. 🍻  

> we would just have to fix the default export and the namespace error.

Like you already said in the comment bellow. I will try to do some work on this issue.

svg.js.d.ts
albohlabs: This is a good starting point to fix the typing.

rdfriedl: I had a lot of free time on my hands so i decided to fix this bug.
If you want to check it out, its over [here](https://github.com/rdfriedl/svg.js/commits/master)

Im not an expert in using git or github so im not sure how we would merge my code into this PR.
any ideas?

albohlabs: Awesome.

I could cherry pick your commits and include them into the PR, if this is ok for you?

rdfriedl: Go right ahead, it's better then me creating a whole new PR

Fuzzyma: You could also create a PR for his PR by forking his fork and merge it there then merge it here. However: Cherrypicking might be faster ^^

rdfriedl:
there is still one problem left to fix.
this code works fine
```typescript
import * as SVG from 'svg.js';
var draw: svgjs.Doc = new SVG.Doc();
document.body.appendChild(draw.node);
```
but this dose not work anymore
```typescript
import SVG from 'svg.js';
var draw: svgjs.Doc = new SVG.Doc();
document.body.appendChild(draw.node);
```
it says that `svg.js` has no default export...
iv spent a lot of time on this and still can figure out to make it work, any ideas?

albohlabs:
This is a core design decision. You can see the discussion and conclusions in [this issue](https://github.com/Microsoft/TypeScript/issues/2242#issuecomment-83694181). 
To solve this we could define a default export in the library (and typings) itself or change the import syntax like [recommended in this issue](https://github.com/Microsoft/TypeScript/issues/3337#issuecomment-107971371).

albohlabs:
[Here is a documentation](https://github.com/localForage/localForage/pull/608/files) change which also address our case.

rdfriedl:
oh i see, i guess typescript modules dont work exactly like ES6 modules 😄
In that case everything looks good

@Fuzzyma can you merge this?

Fuzzyma:
sure thing. Thanks for your work!

Issue #584, opened 2017-02-09
There is a bug when chaining path plot animation like so: path.animate().plot(p1).animate().plot(p2)


RmiTtro:
The bug is that the second call to `plot` interfere with the first call to `plot`. This bug is caused by a combination of two factors. First, the patharray returned by [SVG.Path](https://github.com/svgdotjs/svg.js/blob/516f575cf4ea4921fab4d66fbce3b8dad25e7e66/src/path.js#L13-L19) is cached. Second, the method [plot](https://github.com/svgdotjs/svg.js/blob/516f575cf4ea4921fab4d66fbce3b8dad25e7e66/src/fx.js#L854) of the FX module call the method `morph` on the patharray returned by `SVG.Path` while ideally this should be done in the method `initAnimations`. 

Here is an in-deep explanation of what is happening:

```javascript
// path is an instance of SVG.Path
// p1 and p2 are SVG.PathArray
path.animate().plot(p1).animate().plot(p2)
```
In the first call to `plot`, `this.target().array()` is called which return an `SVG.PathArray`. This patharray is cached in `this.target()._array` where `this.target()` is an instance of `SVG.Path`. After that, `.morph(p1)` is called on that the patharray.

In the second call to `plot`, `this.target().array()` is called which return **the value that was cached previously**, it return the value in `this.target()._array`. After that, `.morph(p2)` is called on that patharray which as for effect of changing the destination of the first plot since it is use the same patharray.

That pretty much sums it up (I just hope my explanation wasn't too bad). As for solving this issue, I think that solving #547 should solve this issue. I will have a look at it!

Fuzzyma:
I am a bit curious about this issue.
When `animate().plot()` is called the first time, `initAnimations()` gets the current `PathArray` and uses that for morphing. However: while animating this path, the cached PathArray is changed because `plot()` is called multiple times while animating the path. So I cant understand how the next call to `animate().plot()` (and therefore the call to `initAnimations()` again) can get the same PathArray than before.
Whenever you call plot with a value, the cached value is updated:
```js
return this.attr('d', (this._array = new SVG.PathArray(p)))
```

so this first cached array should not be equal to the second one.
Where is my mistake?

//EDIT: Oh I know see it. The problem really lays in obtaining the path array. Funny enough there is a comment in the code: 
```js
          // TODO: this is not a clean clone of the array. We may have some unchecked references
          s.animations[i].value = (i == 'plot' ? this.target().array().value : this.target()[i]())
[...]

Issue #582, opened 2017-02-05
Animating TextPath seems broken


dotnetCarpenter:
I'm trying to validate the [TextPath](https://svgdotjs.github.io/elements/textpath/) examples.

Specifically `text.textPath().animate(3000).attr('startOffset', 0.8)` fails because SVG.FX doesn't seem to handle string names but only objects. https://github.com/svgdotjs/svg.js/blob/master/src/fx.js#L761

I also tried another approach, where I try to animate the path (which should use our new basic morph path feature) but since TextPath doesn't implement the `array()` method, this doesn't work either.
```js
var draw = SVG("drawing")
  .viewbox(0, 0, 200, 140)
var text = draw.text(function(add) {
    add.tspan("We go up and down, then we go down, then up again")
})
text
  .path('M10 80 C 40 10, 65 10, 95 80 S 150 150, 180 80')
  .animate(500)
  .plot("M10 80 C 40 150, 65 150, 95 80 S 150 10, 180 80")
```
https://github.com/svgdotjs/svg.js/blob/master/src/fx.js#L854 will fail at this.target().**array()**.morph(p)

Looking at our code coverage, it's no wonder since neither [animating attr](https://coveralls.io/builds/9999433/source?filename=dist%2Fsvg.js#L2034) or [animating plot](https://coveralls.io/builds/9999433/source?filename=dist%2Fsvg.js#L2126) is covered in any of our tests.
We might want to go through the SVG.FX module and add the missing tests in accordance with our documentation. But as usual, time is a huge factor :disappointed: 

I will see if I can find some time to look into this but I haven't dug into this part of SVG.js before and don't have much time.

Any help is greatly appreciated :dancer: !

RmiTtro:
I think I can help you out on this :raised_hand_with_fingers_splayed: . I was planing to work on the issues #547, #552 and #584 which are all in the SVG.FX module so I think I should be able to look this one up too. And since I played a lot with the SVG.FX module I think I'm the right man for the job :metal: !

RmiTtro:
Hi everyone, I managed to work a bit on this issue and here is what I found :

First, I think we will have to replace, in the docs, this `text.textPath().animate(3000).attr('startOffset', 0.8)` by that `text.textPath().animate(3000).attr('startOffset', '80%')`. I found out in the [SVG spec](https://www.w3.org/TR/SVG/text.html#TextPathElementStartOffsetAttribute) that percentage need to be passed as string for this attribute as this line explain :
[...]

Issue #552, opened 2016-11-15
Animation with negative percent position values


skrichten:
When directly using SMIL, I can use negative percent values for x,y.  This can be very handy to do things like this...

https://codepen.io/bradleyboy/pen/lIqag/

When I tried such a value with this library the animation did not play at all.  Any workarounds or other suggestions for achieving this affect would be appreciated.

Fuzzyma:
Please include the relevant code from your example in your question. Also add the code you tried to achieve the effect with this library

skrichten:
Code from the example...

`
<pattern id="pattern" x="0" y="0" width="300%" height="100%" patternUnits="userSpaceOnUse">
      <rect x="0" y="0" width="150%" height="100%" fill="url(#gradient)">
        <animate attributeType="XML"
                 attributeName="x"
                 from="0" to="150%"
                 dur="7s"
                 repeatCount="indefinite"/>
      </rect>
      <rect x="-150%" y="0" width="150%" height="100%" fill="url(#gradient)">
        <animate attributeType="XML"
                 attributeName="x"
                 from="-150%" to="0"
                 dur="7s"
                 repeatCount="indefinite"/>
      </rect>
    </pattern>
`

Code I tried...

`var draw = SVG('drawing'); 

    var gradient = draw.gradient('linear', function(stop) {
        s1 = stop.at(0, '#33235b');
        s2 = stop.at(0.5, '#E97639');
        s3 = stop.at(1, '#33235b');
    });

    var r1, r2;
    var fill = draw.pattern('300%', '100%', function(add) {
      r1 = add.rect('150%', '100%').fill(gradient);
      r2 = add.rect('150%', '100%').fill(gradient);
    });
    fill.attr({patternUnits: 'userSpaceOnUse'});

    r1.attr('x', 0).animate('7s').attr('x', '150%').loop();
    r2.attr('x', '-150%').animate('7s').attr('x', 0).loop();

var text = draw.text('Manifesto').move('50%', '50%').fill(fill);
    text.font({
      size: 70
      , anchor: 'middle'
      , leading: 1
    });`

Sorry about the formatting, I don't know why it is getting messed up.

Anyway, the r1 rectangle animates as expected, but r2 does not.

skrichten:
A related issue... trying to use percentage values in general often doesn't work.  For example...

`var fill = draw.gradient('radial', function(stop) {
        s1 = stop.at(0, '#ffe39b')
[...]

Issue #547, opened 2016-11-04
Sort out leftovers from the old fx module


Fuzzyma:
When I rewrote the fx module, chaining of animations was introduced. As a consequence (and because of other issues), the startpoint of the animated properties isnt evaluated when the call to the function e.g. `dx` or `x` happens but when the animation starts. That makes relative movements possible. And ensures the right startpoint when the next animation starts.

This is done in the `initAnimation` method of the fx module. The code in there is rather ugly because it has to handle the old concept of morphable objects which have a start and a destination value.  
Since the startvalue is overwritten in the `initAnimation` method, we basically dont need to set it anymore.
That leads to code like this:

```js
  x: function(x, relative) {
    // make sure that groups are correctly handled
    if(this.target() instanceof SVG.G){
      this.transform({x:x}, relative)
      return this
    }
    
    // create an EMPTY Number and morph it to destination
    var num = new SVG.Number().morph(x)
    // also make sure that the relative flag is set
    num.relative = relative
    return this.add('x', num)
  }
```

I see no point in creating an empty object which is _really_ filled at another place. So I would like to search for another solution here.
Obtaining the start value from the element is somtimes difficult. For you a bid of madness from the `initAnimation` method:

```js
          // sometimes we get back an object and not the real value, fix this
          if(s.animations[i].value.value){
            s.animations[i].value = s.animations[i].value.value // woooot
          }
```

@wout @dotnetCarpenter @RmiTtro any ideas on how to circumvent this issues are welcome.

**Some more information for the fx module**
Animations are grouped into several groups:

- animations: Everything you can call with a function from svg.js like `move`, `size`...
- attrs: Bare attributes which are animated. This is mostly for attributes which are not covered by any methods from svg.js but can be used for any attribute
- styles: Holds all styles beeing animated
- transforms: Holds all transformations in the right order which are first morphed and then merged into a single matrix. This prevents that e.g. a rotation becomes a flip and so on

When initAnimation is called, all arrays are gone through and the start values are set to the state of the element. For the `animations` array this is a bit tricky because the svg.js method is used to obtain the value. That ofc has many advantages (size, plot, all custom methods) but can also lead to cases where an object is returned instead of a value (e.g. you get back an SVG.Number instead of just a plain Numer). Thats why this is corrected immediatley (see the madness code above).
Maybe we should return always an object. That would makes things less complicated.

Sorry for the long and maybe a bit confusing post. However this issue gnaws at my brain for some time now and I thought its time now :)

[...]

PR #589, opened 2017-02-17
Sort out leftovers from the old fx module

RmiTtro:
After teasing it for all week, I finally managed to finish the pull request I was talking about :raised_hands: . If accepted, it should fixes all these issues: #547, #552, #582 and #584. Most of the ideas behind it come from a discussion I had with @Fuzzyma in issue #547. It also fixes some bugs I noticed while working on it.

Here is a list of the changes and fixes that are made:
* Add a plot and array method to SVG.TextPath (issue #582)
* Make the method plot a getter when no parameter is passed for SVG.Polyline, SVG.Polygon, SVG.Line, SVG.Path (this idea come from issue #547)
* Allow SVG.PointArray to be passed flat array
* Change the regexp SVG.PointArray use to parse string to allow more flexibility in the way spaces and commas can be used to separate points (something like this is now accepted: '1, 2, 3, 4')
* Allow plot to be called with 4 parameters when animating an SVG.Line
* Fix a bug in the plain morphing part of SVG.MorphObj that is in the FX module
* Relative value for SVG.Number are now calculated in its morph method (idea from issue #547)
* Clean up the implementation of the initAnimation method of the FX module (issues #547, #552, #584)


spec/spec/line.js
```diff
@@ -207,14 +216,29 @@ describe('Line', function() {
...
     })
-  })
-  
-})
-
-
-
-
-
+    it('with one argument, change the x1,y1,x2,y2 attributes of the underlying line node', function() {
```
Fuzzyma: Better say "it change x1,y1,... attributes of svg node when string is passed"

spec/spec/path.js
```diff
@@ -184,9 +192,16 @@ describe('Path', function() {
   })
 
   describe('plot()', function() {
-    it('falls back to a single point without an argument', function() {
-      path = draw.path()
-      expect(path.node.getAttribute('d')).toBe('M0 0 ')
+    it('with an argument, change the d attribute of the underlying path node', function() {
```
Fuzzyma: same here - just to make clear what you are passing

spec/spec/polygon.js
```diff
@@ -176,13 +184,19 @@ describe('Polygon', function() {
...
-  
-})
-
-
-
-
 
 
+  describe('plot()', function() {
+    it('with an argument, change the points attribute of the underlying polygon node', function() {
```
Fuzzyma: dito

spec/spec/polyline.js
```diff
@@ -176,13 +184,18 @@ describe('Polyline', function() {
...
-  
-})
-
-
-
-
-
 
+  describe('plot()', function() {
+    it('with an argument, change the points attribute of the underlying polyline node', function() {
```
Fuzzyma: jep

src/fx.js
```diff
@@ -214,45 +214,28 @@ SVG.FX = SVG.invent({
...
-          }
-
-          if(s.animations[i].relative)
-            s.animations[i].destination.value = s.animations[i].destination.value + s.animations[i].value
-
-        }
+        // The condition is because some methods return a normal number instead
+        // of a SVG.Number
+        if(s.animations[i] instanceof SVG.Number)
+          source = new SVG.Number(source)
```
Fuzzyma: you test of SVG.Number is returned and if so, you convert it to SVG.Number _again_. I feel like it should be the other way round. Covert it to SVG.Number when a normal number was detected.

src/line.js
```diff
@@ -34,12 +36,17 @@ SVG.Line = SVG.invent({
...
 , construct: {
     // Create a line element
     line: function(x1, y1, x2, y2) {
-      return this.put(new SVG.Line).plot(x1, y1, x2, y2)
+      var line = this.put(new SVG.Line)
+
+      // make sure plot is called as a setter
+      return (x1 != null) ?
+        line.plot(x1, y1, x2, y2) :
+        line.plot([[0,0], [0,0]])
```
Fuzzyma: It could also be written `return this.put(new SVG.Line).plot(x1 || 0, y1 || 0, x2 || 0, y2 || 0)`. This way you would catch every wrong parameter like `null, undefined` and would convert it to a valid `0`.
You could also go

```js
return this.put(new SVG.Line).plot(x1 != null ? [[x1, y1], [x2, y2]] : [[0,0], [0,0]]))
```
However I guess thats up to you

src/textpath.js
```diff
@@ -28,14 +28,22 @@ SVG.TextPath = SVG.invent({
 
       return this
     }
+    // return the array of the path track element
+  , array: function() {
+      var track = this.track()
+
+      return (track) ? track.array() : null
```
Fuzzyma: why `(track)` and not just `track`? Or even easier `return track || null`?

src/fx.js
RmiTtro: Actually, the test here check if the _destination_ is an SVG.Number (`s.animations[i]` is first set in the method `x`, `y`, etc... of the FX module. If it is, that mean that the _source_ must also be an SVG.Number so it is converted to one.

spec/spec/line.js
RmiTtro: Will do!

src/line.js
RmiTtro: I like what you did with the boolean OR, I need to use that operator more often.

src/textpath.js
RmiTtro: I don't know why I put parens :laughing: on this one I will remove them. I won't be able to use the boolean operator here unfortunately as I have to return the array of the track and not just the track.

src/line.js
RmiTtro: Unfortunately, I can't make this change here. I just tested it and it make some tests fails. I think that is because x1 is not necessarily a number, it can also be an array, a string and a `SVG.PointArray`. 

But, I was able to use the boolean operator in the constructor of `SVG.Path`, `SVG.Polyline` and `SVG.Polygon`. So that comment was still useful :wink: .

RmiTtro:
This last commit makes almost all the changes that were proposed by @Fuzzyma. Only the one related to SVG.Number was left out for the reason I have given above.

Fuzzyma:
Awesome work. Thank you very much!

Issue #594, opened 2017-02-20
Fix TBox


dotnetCarpenter:
The [TBox](http://svgjs.com/geometry/#tbox) example, [`x` and `y` doesn't follow scale](http://jsfiddle.net/qa8y7qp2/2/), even though the docs suggest that they should.

The TBox can be fixed easily:
- get the bbox and apply the scale transformation, resulting in a correctly sized tbox
- get the rbox and use it's cx and cy values to position the tbox at its center

dotnetCarpenter:
@woutf I have a conceptual question about [`TBox`](http://svgjs.com/geometry/?q=TBox#tbox).

Should `.tbox()` return the actual `x`, `y`, `width` etc values and a [matrix](http://svgjs.com/classes/matrix/?q=matrix) object with the transformed values, thus mimicking the actual rendered SVG element or the transformed  `x`, `y`, `width` etc values?

The former requires you to do:
```js
// draw transformed rect
var rect = draw
	.rect(175,175)
	.move(75,75)
	.rotate(20)
	.scale(1, 0.5)
        .transform({
		skewX: 2
	})

// draw tbox
var t = rect.tbox()
draw
	.rect(t.width, t.height)
	.transform(t.matrix)
	.addClass('box')
	.move(t.x, t.y)
```

While the latter requires you to do:
```js
// draw transformed rect
var rect = draw
	.rect(175,175)
	.move(75,75)
	.rotate(20)
	.scale(1, 0.5)
	.transform({
		skewX: 2
	})

// draw tbox
var t = rect.tbox()
draw
	.rect(t.width, t.height)
	.transform({
		rotation: t.rotation
	      , skewX: t.skewX
              // etc. for all other transforms
	})
	.addClass('box')
	.move(t.x, t.y) // not sure how to get the correct x,y values with the transformations
```

The issue with the latter is that, I'm not sure how to get the correct x,y values and that you can not apply the matrix from `.ctm().extract()` without retaining the original x,y values.

The former will return an object with:
![svgjs_tbox1 1](https://cloud.githubusercontent.com/assets/43763/23232426/7eef40ec-f94b-11e6-8299-7f18bfcd0506.png)

As we [discussed previously, it's doubtful that anyone is actually using `tbox()` in its current state](https://github.com/svgdotjs/svgdotjs.github.io/issues/2#issuecomment-279946995), so I'm not too afraid about breaking stuff if we decide to change the return object.

dotnetCarpenter:
[...]

PR #601, opened 2017-02-23
first proposal for TBox

dotnetCarpenter:
This PR is currently missing unit test (tests are failing) and is just a codification of the first proposal in https://github.com/svgdotjs/svg.js/issues/594#issuecomment-281814847


dotnetCarpenter:
`TBox` will be removed in 2.5.0. It doesn't work, and has not worked for while, so we're not breaking anything by removing it.

Ref: https://github.com/svgdotjs/svg.js/issues/594#issuecomment-281953060

PR #602, opened 2017-02-23
remove SVG.TBox

### LINKED ISSUE #594: Fix TBox

dotnetCarpenter:
This PR change `SVG.TBox` to map to `SVG.RBox` with a deprecation warning - telling users to use `SVG.RBox` instead.

~It has a tiny unit-test for this that must be removed in 3.0, which will be evident when we finally remove `SVG.TBox` completely.~

closes #594


spec/spec/boxes.js
```diff
@@ -44,44 +44,8 @@ describe('BBox', function() {
...
-      var box2 = draw.rect(100,100).move(300,400).bbox()
-      var merged = box1.merge(box2)
-      expect(box1).not.toBe(merged)
-      expect(box2).not.toBe(merged)
-      expect(box1.x).toBe(50)
-      expect(box1.y).toBe(50)
-      expect(box2.x).toBe(300)
-      expect(box2.y).toBe(400)
-    })
+  it('should trow an error in 2.x and be removed in 3.x', function() {
```
wout: Small typo here (trow).

dotnetCarpenter:
@Fuzzyma great! I'll make the change

dotnetCarpenter:
@Fuzzyma `new SVG.TBox` will not work with the change though

Fuzzyma:
Yes I know, but who uses `new SVG.TBox` anyway? There is not really a reason to do so :). I approve this!

saivan:
TBox is super useful if you're working in the svg space, rbox and tbox are not the same right? So how do you propose we go from rbox to tbox without getting the user to apply the transformation themselves. Is this documented anywhere?

Fuzzyma:
Tbox was never working consistently. Rbox is doing the exact job. So we don't need Tbox.

saivan:
I just realised the problem was that we need to call `rbox` with the canvas as an argument. This isn't documented anywhere unless I'm mistaken, can we add that somehow?

Fuzzyma:
Yes of course. Feel free to open a pull request to add it to the documentation. We are short on time at the moment :/

saivan:
Yep, I'll try to get to it as soon as I can :)
Although, its not clear to me how your docs are being built.
Any pointers there?

Fuzzyma:
Yes - thats documented here: http://svgjs.com/contributing/documentation/

Issue #505, opened 2016-07-07
size() calculates NaN when bbox width is zero


albohlabs:
I've got a polyline with two points and same x-coordinate.

<img width="517" alt="bildschirmfoto 2016-07-07 um 13 11 39" src="https://cloud.githubusercontent.com/assets/163665/16651480/6423a2e2-4444-11e6-86a1-6bb79ada9169.png">

On resizing the svg element, the `ResizeHandler.resize.event.calc` triggers the  `size` method which calculates `NaN` for the new x. 

<img width="624" alt="bildschirmfoto 2016-07-07 um 13 25 11" src="https://cloud.githubusercontent.com/assets/163665/16651737/42ab0d56-4446-11e6-814f-84bab2331f7e.png">

This would be fixed if there is minimal width of 1px. This could be checked perhaps in the `bbox` method. What do you think?

Fuzzyma:
Its totally fine for a shape to have a width or height of zero. But we have to make sure, that we check for it and don't produce something like division by zero.
Thanks for reporting!

Fuzzyma:
@albohlabs what do you think is the expected behavior when caling lines:

- only scale the possible value (height in your case) and let width untouched
- add half of the width to every point so taht the new line is centered in the size-box you provided

Would be great to have your thoughts about that

albohlabs:
I would also prefer the first one.

PR #625, opened 2017-03-04
fixes SVG.PointArray.size() which created NaN when the bbox of the point array had zero height/width

Fuzzyma:
The following test would fail without this fix:

```js
    it('let coordinates untouched when width/height is zero', function() {
      var array = new SVG.PointArray([10, 10, 10, 20, 10, 30])
      expect(array.size(60, 60).valueOf()).toEqual([[10,10], [10, 40], [10, 70]])

      array = new SVG.PointArray([10, 10, 20, 10, 30, 10])
      expect(array.size(60, 60).valueOf()).toEqual([[10,10], [40, 10], [70, 10]])
    })
```

The same problem exists for paths but in the paths case its much more lines which needs that check so I let it intouched for now


src/pointarray.js
```diff
@@ -87,8 +87,8 @@ SVG.extend(SVG.PointArray, {
 
     // recalculate position of all points according to new size
     for (i = this.value.length - 1; i >= 0; i--) {
-      this.value[i][0] = ((this.value[i][0] - box.x) * width)  / box.width  + box.x
-      this.value[i][1] = ((this.value[i][1] - box.y) * height) / box.height + box.y
+      if(box.width) this.value[i][0] = ((this.value[i][0] - box.x) * width)  / box.width  + box.x
```
dotnetCarpenter: So `.size()` is a noop (no operation) if the shape has no width or height? Should be written in the documentation before we get a `.size()` doesn't work issue filed.

Fuzzyma: Before it produced NaN instead so it wasnt a  noop but a bug. User should know, that if they stretch points which are on one line, nothing will happen. See #505 for the discussion on this and expected behavior.

Fuzzyma:
@dotnetCarpenter I resolved the conflicts. In the master where tests added to the array.js after that PR. Its all corrected now

PR #626, opened 2017-03-05
added SVG.Box, added transform method to boxes, fix rbox

Fuzzyma:
Also:
- workaround for screenCTM with nested svg (FF bug)
- rbox accepts addiotional parameter to transform the box

`el.rbox()` now always returns the correct box in absolute screen coordinates. It also accepts an element as parameter whose screenCTM is used to transform the box.
That way one can easily get the box in the coordinate system he wants.

```js
var rect = draw.rect(100,100)
rect.rbox() // returns box in absolute sreen coordinates
rect.rbox(draw) // same as...
rect.rbox(rect.doc()) // returns the box in coordinates of the docs system
```

The `SVG.RBox` object itself is now relative to the viewport of the root document and does not take scrolling into account.

```js
var rect = draw.rect(100,100)
new SVG.RBox(rect) // returns a box in screen coordinats without scrolling
new SVG.RBox(rect).addOffset() // adds the scrolling
new SVG.Rbox(rect).transform(draw.screenCTM().inverse()) // transforms the box into another system. Same as
rect.rbox(draw)
```

Furtermore `SVG.Box` was added as superclass for the other boxes. It holds the merge and transform method and can be used to just have a simple box:

```js
new SVG.Box(x, y, width, height)
new SVG.Box({x: x, y: y ...})
```

Of course the other boxes accept 4 Parameters / an object now, too.

**Note**: FF has a bug which makes screenCTM on nested svgs useless. A workaround was added to make it consistent with other browsers and much more useful.
https://bugzilla.mozilla.org/show_bug.cgi?id=1344537

I guess thats it. Happy reviewing :)

PS: I didnt put the dist files into the PR to simplify merging


dotnetCarpenter:
@Fuzzyma I just looked it over really quick and so far it looks good. If you want *real* review, you'll have to wait until I get time :(

wout:
Looks good to me too!

PR #634, opened 2017-03-10
unified all boxes

Fuzzyma:
- SVG.Box for everything
- bbox, rbox and viewbox as methods
- boxes can be created with string, array, object, list...
- added helper to simplify boxes code

If you find that the construct methods have to much logic, we could easily use the `SVG.extend` syntax to add these methods to `SVG.Element`. It does not make any difference, though...

The methods are used as before. Everything returns a box now. The zoom in viewbox was removed.


wout:
Yes, please. I'm all for removing as many helper functions as possible.

Fuzzyma:
Oh - guess I abstracted so much that it isn't needed anymore :D. But I
remember that's used at least 2 times. However M next commit will avoid
these, too so yes I will inline it. Thanks for the heads up!

Am 10.03.2017 8:55 nachm. schrieb "Jon Ege Ronnenberg" <
notifications@github.com>:

dotnetCarpenter:
The other place was in the now deleted SVG.TBox ;)

Fuzzyma:
One in merge and one in transform I thought

Am 10.03.2017 9:05 nachm. schrieb "Jon Ege Ronnenberg" <
notifications@github.com>:

> The other place was in the now deleted SVG.TBox ;)
>
> On Fri, Mar 10, 2017 at 8:57 PM, Ulrich-Matthias Schäfer <

dotnetCarpenter:
@Fuzzyma You're right  :) Hmm... then we might need it. But we should expand our boxes tests to include the values `fullbox()` gives so that we can refactor it later without nasty surprises.

Fuzzyma:
As I said I already refactored again and was able to remove all but one use. So we can inline it. I forgot to push those changes. Will see when I can do that.

Fuzzyma:
Ok we now can inline fullbox -  but I just realized, that I added 2 more helpers when writing the code.
Actually I like helpers a lot. Pure functions with a clearly defined task. The make the other code more readable. I thought about making the helper available for other users, too. Maybe by adding them to `SVG.utils` or `SVG` direcly. There are quite a lot helpers which I would potentially use in other code, too (`array_clone`, `is`, `matches`, `camelCase`, `capitalize` ...).
What do you think?

src/boxes.js
```diff
@@ -1,31 +1,35 @@
...
-
-    }
+  create: function(source) {
+    var base = [0,0,0,0]
+    source = typeof source === 'string' ?
+        source.match(SVG.regex.viewbox).map(parseFloat) :
+      Array.isArray(source) ?
+        source :
+      typeof source == 'object' ?
+        [source.left != null ? source.left : source.x, source.top != null ? source.top : source.y, source.width, source.height] :
```
dotnetCarpenter: What is the code coverage here? Do we actually test all of the different constructor signatures? If yes, then GREAT!

src/helpers.js
```diff
@@ -1,3 +1,17 @@
+function isNulledBox(box) {
```
dotnetCarpenter: These helpers are great for reducing the library size. But other helpers are better covered by [Underscore.js](http://underscorejs.org/), [Lodash](https://lodash.com/docs), [functional.js](http://functionaljs.com/) and [Ramda](http://ramdajs.com/) which people already use. We would just duplicate features.

src/helpers.js
```diff
@@ -1,3 +1,17 @@
...
+  return !box.w && !box.h && !box.x && !box.y
+}
+
+function domContains(node) {
```
dotnetCarpenter: But talking about helpers, we could provide a plugin with helpers that are specifically target SVG usage. I would personally have a lot of fun, writing Applicatives and Functors for SVG.

src/regex.js
```diff
@@ -45,6 +45,8 @@ SVG.regex = {
   // split at whitespace and comma
 , delimiter:        /[\s,]+/
 
+, viewbox:          /[+-]?(?:\d+(?:\.\d*)?|\.\d+)(?:e[+-]?\d+)?/gi
```
dotnetCarpenter: Please comment the regex. I can read regex but only 4-5 characters a minute.

src/boxes.js
Fuzzyma: jep all tested

src/helpers.js
Fuzzyma: Most of our helpers are used by ourselves, too. We cant just use another library. So why not make the public when they are there anyway? For the isNullerBox helper I agree that its not much code but it helps a lot to increase readability in the bbox function

src/helpers.js
Fuzzyma: especially this helper would look really ugly when inlined. As I already said: We dont have these helpers for the user in the first place. They are there to do recurring tasks. Making them public is just a plus.

src/regex.js
Fuzzyma: I think we really can replace this one with the normal delimiter regex. This regex matches 4 numbers with optional exponent. But we can just do `split(/[,\s]+/)` instead which is used in all the other cases. So I will remove this one

Fuzzyma:
@dotnetCarpenter what does "LOC" mean?

src/boxes.js
```diff
@@ -46,124 +50,84 @@ SVG.Box = SVG.invent({
...
           clone.remove()
-        }else{
-          box = {
-            x:      element.node.clientLeft
-          , y:      element.node.clientTop
-          , width:  element.node.clientWidth
-          , height: element.node.clientHeight
-          }
+        } catch(e) { 
+          console.warn('Getting a bounding box of this element is not possible')
```
Fuzzyma: I guess we dont need that try catch here because catch wont be ever triggered.
It will be an infinitive loop instead because bbox is called in the try block.
Maybe I can rewrite it to make it better

src/helpers.js
dotnetCarpenter: I do **not** say that we should use another library. I'm saying that we should not duplicate code that other libraries provide, that our users already use.

By all means, create a SVG helper function and re-use them as much as possible. But `matches`, `camelCase` and `capitalize` might be functions that other libraries already provides.

I'm just saying that we should only make helper functions because we need them and it will slim down SVG.js. And that we should be careful not to provide the same functions that other libraries provide.

dotnetCarpenter:
LOC == Lines Of Code

src/helpers.js
Fuzzyma: My second thought about making them availabe was that we can test them. At the moment they are private functions encapsulated in the code. Having them in the SVG obj will make them testable.

I am not sure why it is so bad, to make a camelCase function public api even if other libs have it. Do you think it could fall back on us when we remove our need for that function?
I admit that `matches` does not serve a purpose for the lib. Its a method which you can call on an element already (so its already public). I am completely with you when it comes to helpers which does not bring a benefit but `fullbox` aside I find all our helpers very useful.
However, lets disucss this at another place :)

Issue #631, opened 2017-03-09
Typescript definitions edits: element.stop, container.polyline(), select(): Set, animation.scale()


showdep:
I would suggest a few changes to the TS definition file. My TS code compiles properly against svg.js 2.4.0 after I have made these. All changes are self explanatory (I hope). The scale() thing is discussed in #630.

```
interface Element {
  ...
+  stop(jumpToEnd:boolean,clearQueue:boolean):void;
}

interface Container {
-  polyLine(points: string): PolyLine;
-  polyLine(points: ArrayPoint[]): PolyLine;
+  polyline(points: string): PolyLine;
+  polyline(points: ArrayPoint[]): PolyLine;
}

interface Library {
  ...
-  select(query: string, parent: HTMLElement): Element;
+  select(query: string, parent: HTMLElement): Set;
}

interface Parent {
-  select(query: string): Element;
+  select(query: string): Set;
}

export interface Animation {
  ...
+  scale(w: number, h: number, cx: number, cy:number): Animation;
  ...
}
```

Fuzzyma:
I am pretty sure that the `parent()` does return an element and not a set so I wonder why you wanna change that

dotnetCarpenter:
@showdep do you mind making a PR against 2.5.0?

showdep:
Here it is  #635. I did not walk through all interfaces, just made sure the stuff I am using looks good.

showdep:
Fine with me for now.

PR #635, opened 2017-03-10
#631- Typescript definitions edits: element.stop, container.polyline(), select(): Set, animation.scale()

showdep:


svg.js.d.ts
```diff
@@ -951,7 +952,7 @@ declare namespace svgjs {
 
         rotate(degrees: number, cx?: number, cy?: number): Animation
         skew(x: number, y: number): Animation
-        scale(x: number, y: number): Animation
+        scale(scaleX: number, scaleY: number, cx?: number, cy?: number): Animation
```
Fuzzyma: `scale` may called with one parameter only. So `scaleY` has to be optional, too. Please change that bit and I guess we can merge that. Thanks for your contribution!

dotnetCarpenter:
@showdep I can not push to this PR. Have you unchecked the option?

Anyway. the line `scale(scaleX: number, scaleY: number, cx?: number, cy?: number): Animation` just have to be `scale(scaleX: number, scaleY?: number, cx?: number, cy?: number): Animation`. Then we are all good.

Fuzzyma:
@dotnetCarpenter as far as i know you cant just push to a repository which is owned by someone else if he didnt add you as maintainer explicitely. And thats not often done.

However I agree with the change proposed by @dotnetCarpenter. We can merge that after

Fuzzyma:
@dotnetCarpenter @showdep I allowed myself to do some changes to the ts file. Please doublecheck so we can merge

showdep:
@Fuzzyma 

> I allowed myself to do some changes to the ts file.
> Please doublecheck so we can merge
Fine with me: param names and optional params

Fuzzyma:
@showdep I updated the definitions again. Can you pls check the current master? I have no typescript environment to check if it works. Would be cool if at least the syntax should be correct everywhere :)

PR #640, opened 2017-03-17
a try to combine the transform function of fx and element again

Fuzzyma:
when I rewrote the FX module I was in need to write an own implementation of the `transform` method for `SVG.FX` only. I never liked that - so I tried to combine them again.

The problem here is, that in the combined function every transformation is done by transforming it to a matrix first (using SVG.Transformation - as already in fx). In the normal transform method we still used some easy math at some places (translate and scale). This is ofc faster than creating the matrix, calculating the Matrix which needs to be undone in order to apply an absolute transformation.

I did not compare both functions in the speed aspect - I just wanted to have this work somewhere so it isnt lost. I might add a bench entry to check how it performs.

However, in 3.0 absolute transformations are maybe gone so maybe this isnt needed at all :)


RmiTtro:
Wouldn't you need to make some change to `eachAt` of SVG.FX for this to work?

Fuzzyma:
Actually I didnt touch the fx code at all. I just tried to inline the function of `el.transform` into `fx.transform`. However `fx.transform` works as before because of the `if(!this.is(SVG.FX)) {`

RmiTtro:
You are right, sorry about my dumb question :flushed:. Since some part of your changes looked like what was done in `eachAt`, I guess I got confused!

Fuzzyma:
Yes you are right. I need to do the undo stuff for the transformations on the element as well now. That was only done in eachAt before.

PR #651, opened 2017-04-11
Strict mode

dotnetCarpenter:
Strict mode catch some undeclared variables, which made them global. Not sure if they leaked outside our UMD container, though.

But I think this warrant a patch update to 2.5.2 (we're aligned with Jasmine's version :yin_yang:).

I will check if our benchmark are getting faster/slower after this PR and post the result.


Fuzzyma:
Uhh thats bad - we should have checked that way earlier. Looking forward to standard syntax in 3.0 and finally (annyoing) liniting

dotnetCarpenter:
According to my very non scientific test, we are shaving off some milliseconds for *more* complicated operations. While generating 1000 rects gives us nothing.
First in strict mode:

strict


And lastly in sloppy mode:

nonstrict

dotnetCarpenter:
@RmiTtro @Fuzzyma @wout 
Does anyone oppose a 2.5.2 release today?

PR #656, opened 2017-04-16
fire attr change event

MillerRen:
It is very usefull when we want to do something on attr changed.


Fuzzyma:
Did you check the benchmarks when adding this to the source base? I am afraid firing an event on every attribute update it could impact performance. You could use mutation observers (or mutation events) instead which works exactly like that but natively

PR #661, opened 2017-04-20
add options arguments to `on()`

Fuzzyma:
Chrome replaces the third argument of `addEventListener(event, fn, useCapture)` with `addEventListener(event, fn, options)`. This kinda breaks the web. But thats nothing we should care about since this implementation will still assure standard behavior if no options are passed.

Only the people who will use that feature (like me in svg.panzoom.js) will have to deal with other browsers doing wrong things.

I admit that adding a 5th parameter to `on()` is non optimal - however I dont see another way of doing this. And since this feature will be used less often, iam fine with it.


wout:
As you say, the fifth parameter is not ideal, but I don't see any harm in it.

Issue #568, opened 2017-01-13
<base> tag breaks SVG marker rendering


grzhan:
See http://stackoverflow.com/questions/18259032/using-base-tag-on-a-page-that-contains-svg-marker-elements-fails-to-render-marke

Maybe we need add a parameter in `SVG.Marker` to support fully resolve the marker url.

wout:
We've had this question before. This issue affects all url() references. We'll have to add this to the to-do. Maybe make a utility function that we can use everywhere an url reference is made.

Fuzzyma:
I was searching for that issue and found this one: #36. But actually I thought there was another one.
We decided to let this at it is because it added other problems with references. However, I dont know which one anymore...

wout:
Thought so too and found it!

https://github.com/svgdotjs/svg.filter.js/pull/1

I think we must indeed be careful with exporting. All `url()` references should then be converted on export.

PR #663, opened 2017-04-23
Fix basetag bug

Fuzzyma:
prefix all `url()` references with current location (#568)

the prefix is removed on export and added on import.

TODO:
- [ ] add tests when exporting svg which alrady has a location in fill set e.g. `fill="url(google.de/some.svg#
someId)"` and the same when importing
- [ ] compare performance (which is most likely reduced when importing/exporting svgs)


Fuzzyma:
sry guys - wrong base branch. Opened another PR

PR #664, opened 2017-04-23
Fix basetag bug 

Fuzzyma:
prefix all `url()` references with current location (#568)

the prefix is removed on export and added on import.

TODO:
- [x] add tests when exporting svg which alrady has a location in fill set e.g. `fill="url(google.de/some.svg#
someId)"` and the same when importing
- [ ] compare performance (which is most likely reduced when importing/exporting svgs)


Fuzzyma:
I thought about that, too. But when some genius people start creating the basetag with javascript we are lost again. But one might argue that this is the users fault then because he should know better.

wout:
True. This solution covers all use cases. I think it's good to go then. We might be missing some attributes using an URL, but that's something for later.

Fuzzyma:
Atm every attribute which matches `url(...)` is corrected

Fuzzyma:
But I guess thats not enough. `use` uses `href` which doesnt containa any `url()`. Its just the plain address. I need to check that, too.

src/regex.js
```diff
@@ -39,6 +39,9 @@ SVG.regex = {
   // Test for image url
 , isImage:          /\.(jpg|jpeg|png|gif|svg)(\?[^=]+.*)?/i
 
+  // Test for url reference
+, isUrl:            /url\(([-\w:/]+(?:\.\w+)?)?#([-\w]+)\)/
```
RmiTtro: I think there might be some cases that this regexp is not matching, like `url(https://example.org/absolute/URI/with/absolute/path/to/resource.txt#theid)`. Also, I was wondering if we could use a simpler regexp instead, something like this:  `/url\((.*)#([-\w]+)\)/`. We keep it simple and just check if there is something in front of the id. If there is nothing, then we had the complete url. I don't know if it is worth it to try to detect if the path before the id is absolute or relative.

Fuzzyma: I like to be loose as possible by avoing the `.*` operator. I will go for `/(url\()?([-\w:/.]+)?#([-\w]+)/` which should match everything valid.

Fuzzyma:
Seriously I dont like this...
I added like 150 lines of code and 5 helper functions only to achieve the proper workaround for this few poeple who uses a basetag. I would most likely prefer stating in the docs that people should avoid the basetag and if they have to use it, install a plugin instead.
This is so much overhead for literally nothing...

wout:
Absolutely agree. Let's poor it into a plugin.

RmiTtro:
I 100% agree!

PR #671, opened 2017-04-30
Adds the intended functionality to call animate functions with multiple parameter

Fuzzyma:
While writing the zoom plugin I discovered, that the fx module didnt cover a usecase which was orginally intended to work. You cant pass multilple parameter to methods when animating them.

"Stop", I hear you saying, "`move()` _does_ expect 2 parameters!". Well yes thats true but the implementation of `move()` breaks it down to `x()` and `y()` which both only take one parameter.

So this wasnt possible: `el.animate().zoom(level, point)`.
This PR aims to fix this while adding the possibility to use `static` values which stay the same while animating. In this case: `point`. When you animate zooming you dont want the point to which you zoom to be animated. It stays the same.

So fx now does the following:
- you set `el.animate().method(val1, val2)`
- if `el.method()` returns the same number of arguments (as array) then all arguments are morphed to each other (the behavior everyone expects)
- however when `el.method()` returns _less_ arguments (e.g. `zoom` only retuns the zoomlvl but no point) it will morph the arguments which are there and let the additional arguments untouched (in case of zoom its `point` which is not changed while zooming)
- if `el.method` returns more arguments the whole fx module will explode

Actually I didnt changed that much code but it has quite an impact :)

### DIFF
--- a/src/fx.js
+++ b/src/fx.js
@@ -860,11 +860,11 @@ SVG.extend(SVG.FX, {
   }
   // Add animatable plot
 , plot: function(a, b, c, d) {
-    if(this.target() instanceof SVG.Line && arguments.length == 4) {
+    // Lines can be plotted with 4 arguments
+    if(arguments.length == 4) {
       return this.plot([a, b, c, d])
     }
 
-    // We use arguments here since SVG.Line's plot method can be passed 4 parameters
     return this.add('plot', new (this.target().morphArray)(a))
   }
   // Add leading method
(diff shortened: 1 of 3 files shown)


src/fx.js
```diff
@@ -845,9 +859,13 @@ SVG.extend(SVG.FX, {
     return this
   }
   // Add animatable plot
-, plot: function() {
+, plot: function(a, b, c, d) {
+    if(this.target() instanceof SVG.Line && arguments.length == 4) {
+      return this.plot([a, b, c, d])
+    }
+
     // We use arguments here since SVG.Line's plot method can be passed 4 parameters
```
RmiTtro: I think you can remove this comment

PR #672, opened 2017-05-01
wraps constructor methods so that they accept objects as shorthand

Fuzzyma:
- does not work for many edge cases which are uncovered for now

You can pass an object to any constuctor method which applies all attributes directly on the element. In case you passed an instance of a shape it is cloned.
So this works:
```js   
var rect = canvas.rect({
  width:100,
  height:100,
  x:100,
  y:100,
  fill:'#ccc'
})

canvas.group().rect(rect)
canvas.rect(100,100)
```

But this produces errors:
```js
// arrays are also objects
draw.polygon([200, 200])

// passing a value as object produces problems, too
draw.rect(new SVG.Number(100), 100)
```

This branch does not aims to be merged asap. Its merely something we can start with

### DIFF
--- a/src/svg.js
+++ b/src/svg.js
@@ -47,19 +47,19 @@ SVG.extend = function (modules, methods) {
   }
 }
 
-var wrapFn = function(method, initializer) {
-  return function(o) {
-    if(!(initializer.prototype instanceof SVG.Element))
-      return method.apply(this, [].slice.call(arguments))
+var wrapFn = function (method, initializer) {
+  // not a constructor we want to change
+  if (!(SVG.Shape && initializer.prototype instanceof SVG.Shape)) {
+    return method
+  }
 
-    if(o instanceof initializer) {
-      return o.clone(this)
-    }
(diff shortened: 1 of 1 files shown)


saivan:
I believe this is the code you were talking about right? But you only wanted it to accept attributes instead of method names.

Fuzzyma:
Jep. And its only for attributes. No methods involved here.

However the problems above remain. Ofc you can check if the prototype of passed object is `null` which is true for `{ }` but not for `new Object` or do a bunch of other checks. We have to discuss that. 

And we need to check how performance differs when using this

saivan:
Yes I was going to suggest that too, checking the prototype is probably the way to go!

Fuzzyma:
Its finally done!! Performance impact cant be measured - to much noise. So I think its negible. And it works like a charm. Just try this beautifiul `canvas.rect({width: 100, height:100, x: 100, y: 100, fill: 'blue'})` :)

PR #680, opened 2017-05-20
Add missing type definitions to svg.js.d.ts

rokoroku:
Updated:

* `.attr()` getter method is missing in `Element` and `Animation` 
* `.leading()` getter method is missing in `Text`
* `.lines()` getter method in `Text` is actually return a `Set` of `Tspan`
* `.radius(x: number, y?: number)` method is missing in `Rect`
* `FontData` can accept `size` and `leading` property as `NumberAlias`

### DIFF
--- a/svg.js.d.ts
+++ b/svg.js.d.ts
@@ -62,6 +62,7 @@ declare namespace svgjs {
 
     // attr.js
     interface Element {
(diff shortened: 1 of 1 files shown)


Fuzzyma:
Thanks!

PR #686, opened 2017-06-06
TypeScript fixes and remove unused parameter

maltefl:
Hi, thanks for this awesome lib :) 

The default TypeScript declaration file gave me a few syntax errors when I tried to use it. These are the fixes that made it work for me. I have never used TS before, so this is more or less the result of "hacking by error message", but the changes are fairly simple syntax fixes.

For some of the stuff I touched, at least in the Transformation interface, I'm not sure the declaration exactly matches all use cases of the JS function even now. But I don't know enough about TS or svg.js to be certain.

Going through function definitions, I also found the function spof() in doc.js which had an unused parameter spof (whose type was undeclared in the declaration, hence a syntax error). I removed the parameter.

### DIFF
--- a/src/doc.js
+++ b/src/doc.js
@@ -57,7 +57,7 @@ SVG.Doc = SVG.invent({
     }
     // Fix for possible sub-pixel offset. See:
     // https://bugzilla.mozilla.org/show_bug.cgi?id=608812
-  , spof: function(spof) {
+  , spof: function() {
       var pos = this.node.getScreenCTM()
 
       if (pos)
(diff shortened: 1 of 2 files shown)


Fuzzyma:
Thats odd. Last time I corrected the typescript files I had no error. Maybe I had a newer ts version? 
I have no clue about TS, too. So I did the same as you did.

Anyway. Thanks for the contribution. Its really appreciated!

maltefl:
I'll be working with typescript + svg.js for a leisure project, I'll check back with any tweaks if I find them :)

PR #697, opened 2017-06-27
Set a hostname to allow for remote Safari tests

DylanLacey:
Safari ignores configured proxies when targetting local addresses.

This interferes with how Sauce Connect proxies traffic to remote
hosts like Travis.ci, causing problems actually loading content for
testing.

Setting a hostname obviates this problem, and even after resolution
Safari will use the system proxy for all connections with that host.


dotnetCarpenter:
Hmm.. One concern though is not with the actually PR but with the failing test. According to the travis log the issue is that SAUCE_USERNAME and SAUCE_ACCESS_KEY is not provided but both are defined as environmental variables on the travis settings page.

@DylanLacey Did our credentials change over at sauce labs?

dotnetCarpenter:
Manually merged. github gets confused when you squash the previously commits history.

DylanLacey:
We don't re-generate access keys without a request and I didn't during my testing; Sounds like it was just GH and Travis getting confused?

PR #728, opened 2017-08-03
Use full URL in marker URL

tuongaz:
This will fix issue in some browsers (ex: Safari) the marker is not showing when having `<base href="/">`


Fuzzyma:
Take a look at #664

PR #736, opened 2017-09-14
Allow SVG.Point instances in SVG.PathArray

sharat87:
The SVG.PathArray class currently cannot handle SVG.Point objects in it's input
expression. This changeset add support for that.

### DIFF
--- a/src/patharray.js
+++ b/src/patharray.js
@@ -253,7 +253,7 @@ SVG.extend(SVG.PathArray, {
 
     }else{
       array = array.reduce(function(prev, curr){
-        return [].concat.call(prev, curr)
+        return [].concat.call(prev, curr instanceof SVG.Point ? [curr.x, curr.y] : curr)
       }, [])
     }
 


Fuzzyma:
Sorry for not coming back to this.
Do we actually need this feature? Can you come up with a good usecase for it?

When we implement that I think we should broaden the usecase to objects in general and not only SVG.Points. So the check could be:

```js
typeof curr  === 'object' ? [curr.x, curr.y] : curr
```

But this will fail because Arrays are objects, too. So we have to switch it around:

```js
Array.isArray(curr) ? curr : [curr.x, curr.y]
```

Fuzzyma:
But I just realized: You added this snipped to `PathArray`. That does not make any sense. A PathArray is of the form `[['M', 10, 10], ['L', 30, 30]]`. You cannot create that form out of somethin like `[[10, 10], [30, 30]` (or for objects `[{x: 10, y: 10}, {x: 30, y: 30}].

This PR would be more suitable for PointArrays where this totally makes sense.
Can you clarify?

Issue #805, opened 2018-02-06
Image memory leaks (and possible solution)


kuzmanovicnenad:
First, thanks for awesome lib!

Second, well, it looks like (creating and) removing image is making memory leaks and maybe I've manage to track down the problem...

If you execute next piece of code and keep track of heap memory and events in chrome developer tool (performance tab) you'll find out that for each created and removed image you got +2 events and +1 SVGImageElement detached from DOM.

```
    var draw = SVG('test', 200, 200);

    // Create 100 images
    for(var i = 0; i < 100; i++) {
        var image = draw
            .image('./Testing.jpg', 200, 200);
    }

    // Remove 100 images
    SVG.select('image').each(function () {
        this.off();
        this.remove();
    });

    draw.clear();
    draw.remove();
    draw = null;
```

Those +2 events are 'load' and 'error' events in Image class and (I think) the problem is that events are registered with img element and not this.node element:
```
img  = new window.Image()
SVG.on(img, 'load', function() {
```

Now, when those are to be disposed, SVG.off method is trying to find this.node in handlerMap but can't find it:
```
SVG.off = function(node, event, listener) {
  var index = SVG.handlerMap.indexOf(node) // 'node' is this.node and can't be find because it's registered with img element
   ...
  if(index == -1) return
```

So, after the image is removed the events are not disposed (+2 events) and reference to SVGImageElement still exists in SVG.handlerMap (and SVG.listeners?) (+1 detached DOM element).

What I did is next:
```
img  = new window.Image()
//SVG.on(img, 'load', function() {
SVG.on(this.node, 'load', function() {
```
...and that fixed the mem leaks but I'm not sure if this could cause some different problems...

This also works for me but it's a bit hacky solution:
```
   SVG.handlerMap = [];
   SVG.listeners = [];
```

Fuzzyma:
Good find! Thanks for reporting. 

The reason we bind the load event to the img element instead of this.node is, that (as far as I remember) the load event on svg images is not working as good or does not provide the information needed (width, height...) to properly scale the image.
So you cannot just switch this.

The proper way to fix the memory leak would be to call `SVG.off(img)` in both, loaded handler and error handler. This way we make sure that no event on img remains.

kuzmanovicnenad:
For some reason it's not working in handler but I managed to dispose it by saving this.img and extending remove() function (Image class):
```
...
, extend: {
    remove: function() {
        SVG.Element.prototype.remove.call(this);
[...]

PR #806, opened 2018-02-06
#805 Image memory leak fix

kuzmanovicnenad:
Removing image is not removing "load" and "error" events. (+2 events)
Removing image is not removing reference to image in DOM (+1 SVGImageElement detached dom)
This is fixed by passing this.node when registering "load" and "error" events instad of created img element.
Problem was that this.node could not be found in SVG.handlerMap in "off()" method.


Fuzzyma:
Iam not sure why you make it that complicated. Is just `SVG.off(img)` here not working?

```js
SVG.on(img, 'load', function() {
  SVG.off(img)
  //...
})

SVG.on(img, 'error', function(e){
  SVG.off(img)
  //...
}
```

kuzmanovicnenad:
That was the first thing I tried after your comment but it didn't work.

Fuzzyma:
Actually that should work without any problems. So it needs to be checked why this does not work. No need to work around something which would fix the problem in an elegant way.
I really wonder why this does not work. Have to doublecheck that myself

kuzmanovicnenad:
Nice!
Can you please make a new release with the fix?

Fuzzyma:
https://github.com/svgdotjs/svg.js/releases/tag/2.6.4

Issue #807, opened 2018-02-07
Save event handler on node or instance itself


Fuzzyma:
#### Problem

Issue #805 revealed a memory leak on our library. It can be resolved easily but it made me thinking.
When someone binds an event to an element, the element gets saved in an array so we have a connection from element to handler. However when the element is removed from the dom, the array is not cleaned up.

@kuzmanovicnenad mentioned that it would be good to remove all event handlers when we remove the element from the dom. However, this approach has the problem that dom nodes can be reattached to the dom. A simple `front()` call deletes the node and reattaches it at the end of the node list.
So this solution is not so great.

#### Solution
Thats why I propose the solution, that the event handlers are simply not saved in an array anymore but on the element itself. We could have a property `_events` on any instance where we can save this sort of stuff. But this holds a problem, too. We allow binding of events on normal HTML nodes as well so we would need to save the handlers on the node which means: Altering the native API - baah.

But I dont see a better solution to prevent memory leaks. Because: When all references to a node are dropped and the node is not in the dom anymore, everything saved on the node is cleaned automatically. One thing less to think about.

@saivan @wout @kuzmanovicnenad any thoughts on this?

saivan:
I'm not so clear as to why what you're proposing (which I think is quite sensible btw), is going to require a modification to the native API, can you explain some more?

Fuzzyma:
In case someone wants to bind events to html nodes like this:

```js
var div = document.getElementById('SomeDiv')

SVG.on(div, 'eventname', fn)
```

it would save the handlers on the html node:

```js
div.events['eventname'] = fn
```

And in my opinion any property change on a native element which is not covered by the spec is a modification of the native api.

This is no modification: `node.id = 'newId'`
But this is: `node.someRandomName = 'foo'`

kuzmanovicnenad:
appendChild() won't remove events from element so front() method is safe (unless I'm missing something in your code..)

You don't have to remove events on remove() if off() is working properly (but I would expect it from methods name), and if lib is used properly. 
I think it's developers responsibility to use remove() or clear() methods for removing elements and not to remove it with $('rect').remove() or whatever. But you can't stop them, can you?

I don't know if it's possible but can you just attach native event handlers to elements? 
This would fix the problem above and need for a potential mem leak array.

Regarding this:
```
div.events['eventname'] = fn
```
It will make a problem if:
```
SVG.on(div, 'click', fn1)
SVG.on(div, 'click', fn2)
```
[...]

Issue #550, opened 2016-11-13
Make events cancelable


Fuzzyma:
On default the events created by `fire` are not cancelable. That means a call to `preventDefault()` on them will not have any effect. I just tried to add an option for this to the `fire` method but stumbled over another problem.
To check if an event was `defaultPrevented` you look into said argument. When its true, someone called `preventDefault()` on it. However: To check this, you need the event object which is capsulated from the user with the fire method. We also cant change the return value from `fire` without breaking the api.

Ofc you can create your event outside of `fire` and pass it to the method but thats replicated code all day long. So I thought about a helper function like this:

```js
SVG.getEvent = function(name, data, cancelable){
    return new CustomEvent(name, {details: data, cancelable: cancelable})
}
```

But that doesnt really save much charakters...
Would be cool after all to have the `fire` return the event so you could just do:

```js
if(el.fire(name, data, true).defaultPrevented)
    return
```

However: Any thoughts?

Fuzzyma:
@wout @dotnetCarpenter @RmiTtro I would like to hear your ideas on this.
I really would like to have that into 2.5 but I am not sure how to implement it

Fuzzyma:
Ok so I have 5 different proposals sorted from (imo) bad to better. 

1. **The ugly hack**
```js
  // Fire given event
, fire: function(event, data) {

    // Dispatch event
    if(event instanceof Event){
        this.node.dispatchEvent(event)
    }else{
        this.node.dispatchEvent(event = new CustomEvent(event, {detail:data}))
    }
    
    this.event = function() {
      delete this.event
      return event
    }.bind(this)

    return this
  }
```

Whever an event is fired, a function gets added to this, which returns the event. The function is deleted after being used (uaah - never wrote such a big hack).

2. **The C way**
```js
, fire: function(event, data, returnObj) {

    // Dispatch event
    if(event instanceof Event){
        this.node.dispatchEvent(event)
    }else{
        this.node.dispatchEvent(event = new CustomEvent(event, {detail:data}))
    }
    
    if(returnObj) returnObj.event = event

    return this
  }
```

Pass an object to the function. Add the event to that object.


3. **The confusing return**
```js
, fire: function(event, data, returnEvent) {

    // Dispatch event
    if(event instanceof Event){
        this.node.dispatchEvent(event)
    }else{
        this.node.dispatchEvent(event = new CustomEvent(event, {detail:data}))
    }
    
    if(returnEvent) return event

    return this
  }
```

[...]

PR #808, opened 2018-02-11
proposal for #807, includes #550, `SVG.off()` multiple events and option argument

Fuzzyma:
This PR is Work in Progress.

It moves the event handlers on the node as discussed in #807. It also changes #550 so that we now have `fire()` and `dispatch()` as two seperate methods.

Furthermore its now possible to remove multiple elements at once when calling `SVG.off()`. (Binding was already possible). It also adds the option argument to `SVG.off()` which is needed there.

I made this PR to see which changes are required to make the proposal work. Its not that much overhead to manage HTMLNodes.
Comments are welcome!

### DIFF
--- a/spec/spec/event.js
+++ b/spec/spec/event.js
@@ -10,7 +10,8 @@ describe('Event', function() {
 
   beforeEach(function() {
     rect = draw.rect(100, 100)
-    spyOn(SVG,'on').and.callThrough()
+    spyOn(SVG, 'on').and.callThrough()
+    spyOn(rect, 'dispatch').and.callThrough()
   })
 
   afterEach(function() {
@@ -25,8 +26,8 @@ describe('Event', function() {
     , 'mouseover'
     , 'mouseout'
     , 'mousemove'
-    // , 'mouseenter' -> not supported by IE
-    // , 'mouseleave' -> not supported by IE
+    , 'mouseenter'
+    , 'mouseleave'
     ].forEach(function(event) {
       describe(event+'()', function() {
         it('calls `on()` with '+event+' as event', function() {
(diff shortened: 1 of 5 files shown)


spec/spec/event.js
```diff
@@ -71,39 +72,32 @@ describe('Event', function() {
...
       rect2.on('event', action)
       rect3.on('event', function(){ butter = 'melting' })
       rect3.on('event', action)
 
-      expect(Object.keys(SVG.listeners[SVG.handlerMap.indexOf(rect.node)]['event']['*']).length).toBe(1)  // 1 listener on rect
-      expect(Object.keys(SVG.listeners[SVG.handlerMap.indexOf(rect2.node)]['event']['*']).length).toBe(1) // 1 listener on rect2
-      expect(Object.keys(SVG.listeners[SVG.handlerMap.indexOf(rect3.node)]['event']['*']).length).toBe(2) // 2 listener on rect3
-
-      expect(SVG.listeners.length).toBe(listenerCnt + 3)                                                  // added listeners on 3 different elements
+      expect(Object.keys(rect.events['event']['*']).length).toBe(1)  // 1 listener on rect
```
saivan: Whats this star doing?

Fuzzyma: The star is a wildcard namespace. So every event not bound to a particular namespace will get saved under the wildcard namespace

saivan: Oh okay, sounds fair to me.

saivan: Maybe though, it should be clearer? Maybe something like 'default'?

Fuzzyma: I just wanted to take something which wouldnt be used by the user on accident

saivan:
Go ahead and merge this and close its relevant issues if you're happy with it :)

Issue #809, opened 2018-02-11
Merge SVG.Doc and SVG.Nested


Fuzzyma:
When I reworked svg.js to support stuff like `SVG('<svg><rect ...></svg>').addTo('body')` I came across a problem with SVG.Nested and SVG.Doc.

An example:
```js
var foo = SVG('<svg><rect ...></svg>') // SVG.Doc

foo.addTo('#otherSVG') // ok - now it needs to be a nested element

foo.addTo('body') // and now it has to be a Doc again
```

I implemented a few hacks to convert SVG.Doc and SVG.Nested back and forth to make sure its always the right object but that wasnt clean. Furthermore when someone moves the nodes by hand (with native code), I couldnt cover it.

And today it struck me like a lighning. Here is my prosal:

**Introduce an `isDoc()` method:**
```js
isDoc: function() {
  return this.node.parentNode == null || this.node.parentNode.nodeName == '#document' || this.node.parentNode != SVGElement
}
```

We say, that we always handle a node as a doc, when it either

- has no parent
- has a parent which is the document itself (standalone svg)
- the parent is not an SVGElemenet (therefore HTML)

**make `doc()` check if element is a doc or not**
```js
doc: function() {
  if(this.isDoc()) return this
  var parent
  while(parent = this.parent(SVG.Doc)) {
    if(parent.isDoc()) return parent
  }
  
  throw new Error('This should never be reached')
}
```

The `doc()` method for SVG.Element could be then rewritten to:

```js
doc: function() {
  return this.parent(SVG.Doc).doc()
}
```

The big advantage of this is, that we never need to care if the element is SVG.Doc or SVG.Nested. All methods on SVG.Doc like `defs()` and `namespaces()` will redirected to the real doc with:

```js
defs: function() {
  if(!this.isDoc()) return this.doc().defs()

  // do stuff
}

wout:
Looks great to me!

Fuzzyma:
The only edge case which remains is, when someone creates a doc with the SVG constructor and adds it to another SVG, the namespace gets added anyway. But I really dont know how to catch this without again hacking into `put` and `add` to remove it there.
Btw: I think `isRoot()` would fit better

Fuzzyma:
MERGED!!!

PR #811, opened 2018-02-11
merge SVG.Doc and SVG.Nested. Add isRoot() method

Fuzzyma:
Sample implementation of #809


saivan:
Yep I like it as well, no complaints here!

src/doc.js
```diff
@@ -12,15 +12,30 @@ SVG.Doc = SVG.invent({
...
+    }
+  , doc: function() {
+    if(this.isRoot()) return this
+
+    var parent
+    while(parent = this.parent(SVG.Doc)) {
+      if(parent.isRoot()) return parent
+    }
+
+    throw new Error('This should never be reached')
```
saivan: Maybe remove?

saivan: Or return null

Issue #645, opened 2017-03-26
Changes to SVG.Set


wout:
### Precis
`SVG.Set` is incomplete and outdated and it is not consistently returned throughout the library. It should be either updated or removed.

### Problem
1. `SVG.Set` is missing some methods like `text()` to name one, but there are many more
2. The `inherit()` method is a bad approach
3. Members are not easily accessible (`set[3]` for example; should always go through `set.get(3)`)

### Proposal
- [ ] Remove the `inherit()` concept altogether
- [ ] Make `each()` accept a key/value pair (e.g. `set.each('fill', '#f06')` or `set.each('attr', 'fill', '#f06')`)
- [ ] Remove the internal `members` array and store members in indecis directly on the object
- [ ] Make `length` a property instead of a method (more array-like)
- [ ] Add the `collect()` method for collecting attributes (e.g. `set.collect('fill')` or `set.collect(function() { return this.toHex() })`; may be called `map()` too
- [ ] Return, where possible, an instance of `SVG.Set` rather than an array
- [ ] Add `toArray()` method to convert set to an array (or use `valueOf()` for that purpose)

Fuzzyma:
And `remove` should remove by index and not by element because you could have one element twice in a set

saivan:
Then adding something like map would also make sense.

saivan:
If a user intends to use a set, they could just use our list like this:

```
let list = // output from some SVG.js Function
let set = new Set(...list)
```

wout:
@Fuzzyma What are the scenarios for having the same element twice in a `Set`?

Fuzzyma:
I often used set more as an array (I could have used an array instead) and added elements to it myself with `add()`. Nobody stops you from adding one element two times.

I had a talk with saivan yesterday and he mentioned, that he really likes the old set beside its flaws.
It really costs to overcome that - but maybe we should let it have its old syntax...

Then we would just improve it by:

 - calling SVG.Set.inherit only once at the very end of the dist file. This we can get rid of it in the extend method which is a huge plus. Downside: Plugins need to call it themselves.
- adding the features discussed above (index access, each-improvement...)
- rename it to SVG.List to make intention more clear.

I am REALLY not a fan of the inherit code. Put people seems to really like it. And I understand thats easier to write

```js
set.animate(200,200).move(200,200)
```

instead of

```js
set.each('animate', 200, 200).each('move', 200, 200)
```

and in the end I know no better solution

[...]

PR #812, opened 2018-02-11
Revive Svg.Set

Fuzzyma:
This PR brings back SVG.Set which was removed earlier.
It is now implemented as what is called an array-like object. This is nothing more than an object with numeral properties and a length property.
This has the benefit that we can use all array methods on it and it will just work.
It is designed after the proposal #645. The map/collect method is missing atm but can be easily added later.

### DIFF
--- a/src/set.js
+++ b/src/set.js
@@ -4,89 +4,150 @@ SVG.Set = SVG.invent({
     for (var i = members.length - 1; i >= 0; --i) {
       this[i] = members[i]
     }
-  }
+  },
   // Add class methods
-, extend: {
+  extend: {
     // Add element to set
-    add: function() {
+    add: function () {
       Array.prototype.push.apply(this, arguments)
       return this
-    }
+    },
     // Remove element from set
-  , remove: function(element) {
+    remove: function (element) {
       var i = this.index(element)
 
       // remove given child
-      if (i > -1)
+      if (i > -1) {
         Array.prototype.splice.call(this, i, 1)
+      }
 
       return this
-    }
-  , each: function (block, value) {
-      if (typeof block == 'function') {
-        for (var i = 0, il = this.members.length; i < il; i++) {
-          block.call(this[i], i, this)
-        }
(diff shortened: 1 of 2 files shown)


saivan:
Is there a reason not to just use the built in es6 `Set` and include a polyfill?

src/set.js
```diff
@@ -0,0 +1,92 @@
...
+      // return an empty box of there are no members
+      if (this.length == 0)
+        return new SVG.Box()
+
+      // get the first rbox and update the target bbox
+      var rbox = this[0].rbox(this[0].doc())
+
+      this.each(function() {
+        // user rbox for correct position and visual representation
+        rbox = rbox.merge(this.rbox(this.doc()))
```
saivan: Why not store a reference to this.doc() above each?

src/set.js
```diff
@@ -0,0 +1,92 @@
...
+  , remove: function(element) {
+      var i = this.index(element)
+
+      // remove given child
+      if (i > -1)
+        Array.prototype.splice.call(this, i, 1)
+
+      return this
+    }
+  , each: function (block, value) {
```
saivan: Its unclear to me what block should be if it isn't a function?
Can you clarify?

src/set.js
Fuzzyma: Good question. Will change that!

src/set.js
Fuzzyma: it can be a property of an element, too. I will rename it to property to make the intention more clear

saivan:
Does this need changes now because of the discussion we had over the phone?

Fuzzyma:
Jeah the question is if we keep the inherit stuff. It sounded like you had a strong opinion about its usefulness. And I agree that the each solution can be quite ugly.
So we might change it

saivan:
Yep, I mean; I think it is very useful, as we discussed; but perhaps we should discuss some more once we have better solutions to keep it 😃

PR #820, opened 2018-02-22
Fix typings for missing element.js method

niceilm:
I found missing methods(parent(), doc()) for Element

Thanks


Fuzzyma:
why was #819 closed?

niceilm:
Since the return value of the parent method for the Doc type is HTMLElement, the first type definition(#829) has an error, so I corrected it and uploaded it again.

Fuzzyma:
hehe - next time just push the change to the same branch and it will be fine :)

PR #829, opened 2018-03-03
786 new transformations

saivan:
The new transformations are complete. Although they don't work with the fx module just yet.


gulpfile.js
```diff
@@ -97,8 +97,8 @@ gulp.task('unify', ['clean'], function () {
     .pipe(standard())
     .pipe(standard.reporter('default', {
       showRuleNames: true,
-      breakOnError: true,
-      quiet: true
+      breakOnError: process.argv[2] !== "--dont-break",
+      quiet: true,
     }))
```
Fuzzyma: can we make a gulp task for linting only? So we can do `gulp lint`?

package.json
```diff
@@ -56,7 +56,8 @@
   "scripts": {
     "build": "gulp",
     "build:test": "gulp unify",
-    "test": "karma start .config/karma.conf.js --single-run",
+    "build:dev": "gulp --dont-break",
```
Fuzzyma: I feel like build:test and build: dev are the same. We should merge them

spec/SpecRunner.html
```diff
@@ -77,8 +77,8 @@
   <script src="spec/marker.js"></script>
   <script src="spec/mask.js"></script>
   <script src="spec/matrix.js"></script>
-  <script src="spec/memory.js"></script>
   <script src="spec/number.js"></script>
+  <script src="spec/nested.js"></script>
```
Fuzzyma: this has to go. nested was removed

src/matrix.js
```diff
@@ -24,53 +24,150 @@ SVG.Matrix = SVG.invent({
...
+      var ox = o.origin && o.origin.length ? o.origin[0] : o.ox || 0
+      var oy = o.origin && o.origin.length ? o.origin[1] : o.oy || 0
+      var px = o.position && o.position.length ? o.position[0] : o.px
+      var py = o.position && o.position.length ? o.position[1] : o.py
+      var tx = o.translate && o.translate.length ? o.translate[0] : o.tx || 0
+      var ty = o.translate && o.translate.length ? o.translate[1] : o.ty || 0
+      var rx = o.relative && o.relative.length ? o.relative[0] : o.rx || 0
+      var ry = o.relative && o.relative.length ? o.relative[1] : o.ry || 0
+      var currentTransform = new SVG.Matrix(this)
+
```
Fuzzyma: what happens when passing an actual point to origin, position and so on?
I would suggest using a point to get what you need from the input. Something along these lines:

```js
var p = new SVG.Point(o.ox == null ? origin : o.ox, o.oy)
var ox = p.x
var oy = o.y
```

That syntax should allow for:

- `origin: [10, 20]`
- `origin: {x: 10, y: 20}`
- `ox: 10, ox: 20`
- `undefined`

src/matrix.js
```diff
@@ -24,53 +24,150 @@ SVG.Matrix = SVG.invent({
...
+        // Return the affine parameters
+        scaleX: sx,
+        scaleY: sy,
+        shear: lam,
+        rotate: theta,
+        translateX: e,
+        translateY: f,
+
+        // Return the matrix parameters
+        matrix: new SVG.Matrix(this),
```
Fuzzyma: I actually think we can skip this one when we have a-f in this object anyway. The object you return back is already a matrix-like object because of a-f. So you can pass it in where a matrix is needed directly. And if you want a clean matrix you can just do `new SVG.Matrix(el.transform())`

src/matrix.js
```diff
@@ -82,21 +179,47 @@ SVG.Matrix = SVG.invent({
...
+      var a = l.a * r.a + l.c * r.b
+      var b = l.b * r.a + l.d * r.b
+      var c = l.a * r.c + l.c * r.d
+      var d = l.b * r.c + l.d * r.d
+      var e = l.e + l.a * r.e + l.c * r.f
+      var f = l.f + l.b * r.e + l.d * r.f
+
+      // Form the matrix and return it
+      var product = new SVG.Matrix(a, b, c, d, e, f)
+      return product
```
Fuzzyma: So I guess native is completely gone? :D

Fuzzyma: I actually wonder which is faster

Fuzzyma: I think it was done to reduce filesize and potential errors

src/matrix.js
```diff
@@ -82,21 +179,47 @@ SVG.Matrix = SVG.invent({
...
     },
+
+    lmultiply: function (matrix) {
+      var l = parseMatrix(matrix)
+      return l.multiply(this)
+    },
```
Fuzzyma: you could go `return parseMatrix(matrix).multiply(this)` directly if you like.
Also: What is parseMatrix doing? Shouldnt be a `new SVG.Matrix(matrix)` be sufficiant?

src/matrix.js
```diff
@@ -82,21 +179,47 @@ SVG.Matrix = SVG.invent({
...
+
     // Inverses matrix
     inverse: function () {
       return new SVG.Matrix(this.native().inverse())
     },
```
Fuzzyma: oh its not gone :)

src/matrix.js
```diff
@@ -105,22 +228,38 @@ SVG.Matrix = SVG.invent({
         y = x
       }
 
-      return this.around(cx, cy, new SVG.Matrix(x, 0, 0, y, 0, 0))
+      // Rotate the current matrix
```
Fuzzyma: Should be "Scale the current matrix" I think

src/sugar.js
```diff
@@ -34,47 +34,73 @@ var sugar = {
 })
 
 SVG.extend([SVG.Element, SVG.FX], {
+  // Let the user set the matrix directly
+  matrix: function (mat, b, c, d, e, f) {
+    var matrix = new SVG.Matrix(arguments.length > 1 ? [mat, b, c, d, e, f] : mat)
```
Fuzzyma: The good thing about the Matrix constructor is: you can just go `new SVG.Matrix(mat, b, c, d, e, f, )` and it will work. The constructor will see when mat is a matrix and use it

Fuzzyma: however we need to check if at least ONE argument was given. Because if not, we have to return the matrix. Atm this code will fail when calling it without argument.

So maybe like this?

```js
if(mat == null) return new SVG.Matrix(this) // or this.matrixify()

return this.attr('transform', new SVG.Matrix(mat, b, c, d, e, f))
```

Looks good enough I think :)

src/group.js
```diff
@@ -9,11 +9,11 @@ SVG.G = SVG.invent({
   extend: {
     // Move over x-axis
     x: function (x) {
-      return x == null ? this.transform('x') : this.transform({ x: x - this.x() }, true)
+      return x == null ? this.transform().e : this.translate(x - this.gbox().x, 0)
```
Fuzzyma: Did you find a better solution for this? You mentioned that we could hget rid of gbox somehow

CHANGELOG.md
```diff
@@ -63,7 +65,9 @@ The document follows the conventions described in [“Keep a CHANGELOG”](http:
 - default constructor now has an optional `node` argument which is used to consruct the object e.g. `new SVG.Rect(rectNode)`
 - SVG.Elements constructor now tries to import svgjs:data from the node
 - `SVG.on()` calls the listener in the context of the passed object. el.on always uses the svg.js object as context
-- `SVG.on()/off()` and `el.on()/off()` now accepts multiple comma or space seperated events e.g. "mousedown, foo bar" (#727)
+- `SVG.on()/off()` and `el.on()/off()` now accepts multiple comma or space separated events e.g. "mousedown, foo bar" (#727)
+- Matrices now apply transformations like `scale`, `translate`, etc... by left multiplying them to simplify transformations
+- Transforming an element is now much simpler
```
Fuzzyma: maybe better write something like "The way `transform()` works had been completely reworked. See the docs for further reference"

package.json
saivan: gulp --dont-break is mainly for the linter. But I am all for killing `build:test`. I don't think its necessary anymore

spec/SpecRunner.html
saivan: Oh okay. Weird it didn't error on that. No issues

saivan: done.

src/matrix.js
saivan: Good point. Done!

src/matrix.js
saivan: Yea its not really necessary to do this. I could put it back. But it just avoids having to parse both matrices twice.

src/matrix.js
saivan: I'm not sure. I just kept parseMatrix because it was already there. I didn't add it? What is parseMatrix doing :P I think it also allows for strings. Does the method you mentioned work with strings?

src/matrix.js
saivan: inverse? Why would we kill inverse :o Lol

src/matrix.js
saivan: good catch!

src/sugar.js
saivan: Looks good to me
I'll change that.

CHANGELOG.md
saivan: Did it!

spec/SpecRunner.html
Fuzzyma: jeah its just a not found script file. it gives a network error and that was it^^

src/matrix.js
Fuzzyma: yes it can handle strings. parseMatrix actually does nothing:

```js
// Parse matrix if required
function parseMatrix (matrix) {
  if (!(matrix instanceof SVG.Matrix)) {
    matrix = new SVG.Matrix(matrix)
  }

  return matrix
}
```

Ok maybe its better. It does not clones teh matrix. its a bit faster. so maybe we should always use that one :D

src/matrix.js
Fuzzyma: na - native is not gone :D

gulpfile.js
saivan: Done!

src/matrix.js
saivan: Done

src/matrix.js
saivan: Done!

src/group.js
saivan: Yep, testing now!

src/matrix.js
```diff
@@ -1,17 +1,17 @@
...
+/* global abcdef, arrayToMatrix, closeEnough */
 
 SVG.Matrix = SVG.invent({
   // Initialize
   create: function (source) {
     var base = arrayToMatrix([1, 0, 0, 1, 0, 0])
     var i
 
-    // ensure source as object
+    // ensure source as object// ens
```
Fuzzyma: loos weired to me

src/sugar.js
```diff
@@ -83,7 +88,7 @@ SVG.extend([SVG.Element, SVG.FX], {
       : (direction === 'y') ? [0, around]
       : isFinite(direction) ? [direction, direction]
       : [0, 0]
-    this.transform({flip: directionString, origin: origin}, true)
+    this.transform({flip: directionString, ox: origin[0], oy: origin[1]}, true)
```
Fuzzyma: cant you pass just origin here?

dist/svg.js
```diff
@@ -3454,22 +3454,6 @@ SVG.G = SVG.invent({
...
-    },
-    // Move by center over x-axis
-    cx: function (x) {
-      return x == null ? this.gbox().cx : this.x(x - this.gbox().width / 2)
-    },
-    // Move by center over y-axis
-    cy: function (y) {
-      return y == null ? this.gbox().cy : this.y(y - this.gbox().height / 2)
-    },
     gbox: function () {
```
Fuzzyma: can be removed now

src/sugar.js
saivan: yes, but I was being lazy: I'd have to change all the specs :P LOL

PR #830, opened 2018-03-18
Typescript: X & Y Properties on Point

FrankSandqvist:
Or should I not access these directly?


Fuzzyma:
Without accessing x and y that object would be kinda pointless (pun not intended^^) :D

FrankSandqvist:
Haha, yes I figured. :)

PR #831, opened 2018-03-20
WIP: type definitions fixed for selector.js + added click method for …

mpro7:
…set.js

While testing today the svg.js library with typescript, I have found two possible bugs / missing things, which described in [this post](https://stackoverflow.com/questions/49387057/using-svg-select-method-with-class-selector-of-the-svg-js-library-doesnt-work). First, the `select()` method didn't work like mentioned in [documentation](http://svgjs.com/referencing/#svg-select):

Second, for the result of selection, there was no `click()` method:

I have pushed a possible solution and my first ever pull request to open source project on GitHub, so I hope to do not mess up with anything. Also since I haven't found any contributing instructions in this repository, I've tried to follow [the suggested guide](https://egghead.io/courses/how-to-contribute-to-an-open-source-project-on-github). Let me know if I missed something like test etc (I did run `npm run test`), this would be useful knowledge for me for the future contributions.

### DIFF
--- a/svg.js.d.ts
+++ b/svg.js.d.ts
@@ -754,10 +754,10 @@ declare namespace svgjs {
     // selector.js
     interface Library {
         get(id: string): Element;
-        select(query: string, parent: HTMLElement): Set;
+        select(query: string): Set;
     }
     interface Parent {
-        select(query: string): Set;
+        select(query: string, parent: HTMLElement): Set;
     }
 
     // set.js
@@ -775,6 +775,7 @@ declare namespace svgjs {
         last(): Element;
         valueOf(): Element[];
         bbox(): BBox;
+        click(cb: Function): Set;
     }
     interface Container { set(members?: Element[]): Set; }
     interface Library { Set: Set; }
--- a/svg.js.d.ts
+++ b/svg.js.d.ts
@@ -754,10 +754,10 @@ declare namespace svgjs {
     // selector.js
     interface Library {
         get(id: string): Element;
-        select(query: string): Set;
+        select(query: string, parent?: HTMLElement): Set;
     }
     interface Parent {
-        select(query: string, parent: HTMLElement): Set;
+        select(query: string): Set;
     }
 
     // set.js


saivan:
It's awesome to see these contributions! Perhaps the main issue is that we are currently overhauling SVG.Set in the 3.0.0 Branch. It's a little unstable at the moment, but we plan to get to it pretty soon. I don't personally use typescript too often myself, so I'll wait to see if anybody has any issues with this before merging it. But these definitions might need to be redone for 3.0.0

Fuzzyma:
Thanks for your contribution. Unfortunately nobody of the maintainer does know typescript very well so the d.ts file is pretty much maintained by the community. Thats why contributions such as yours are important.

I will add comments to the code where I have questions!

svg.js.d.ts
```diff
@@ -754,10 +754,10 @@ declare namespace svgjs {
     // selector.js
     interface Library {
         get(id: string): Element;
-        select(query: string, parent: HTMLElement): Set;
+        select(query: string): Set;
```
Fuzzyma: `SVG.select()` has the method signature `select(query, parentNode)` so removing the second parameter will most likely not work. But it is optional so adding a simple `?` might help instead

svg.js.d.ts
```diff
@@ -754,10 +754,10 @@ declare namespace svgjs {
     // selector.js
     interface Library {
         get(id: string): Element;
-        select(query: string, parent: HTMLElement): Set;
+        select(query: string): Set;
     }
     interface Parent {
-        select(query: string): Set;
+        select(query: string, parent: HTMLElement): Set;
```
Fuzzyma: On the other hand `SVG.Parent.select()` does NOT have a second parameter. The parent is already defined because you called the method on a parent (like in `group.select(query)`. Are you sure you didnt mix up something here?

svg.js.d.ts
mpro7: I'm not the typescript master either (yet) ;) and it seems that you are right. The `?` will do the job, so the whole line should look like this: `select(query: string, parent?: HTMLElement): Set;'. So basically to make it works, it's enough to make this parameter as optional as it stands in the documentation.

svg.js.d.ts
mpro7: Like in above, this change is not necessary, so will bring it back and try to push changes in the evening.

mpro7:
Sorry for the delay, just have got unexpected travels. I've followed your suggestions, applied requested changes and pushed them out.

PR #845, opened 2018-04-04
Fix for animate() path d attribute

Folkenberg:
.. where tags like 'M', 'A', 'L' etc resulted in NaN entries in the scenario.


Folkenberg:
Hmm.. I guess I'm not well enough into the code to tell why animating the Path attribute uses SVG.Array and not SVG.PathArray. 
However, I can tell where and why it goes wrong in SVG.Array. Below is a small sample for reproduction. When switching to the official public svg.js, the path arrays gets parsed and morphed in SVG.Array and results in a bunches of NaNs.



<html>
<head>
    <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
</head>
<body>
    
    

    
    <script src="https://rawgit.com/Folkenberg/svg.js/master/dist/svg.js"></script>

    <div class="container center">
    <svg id="svgRoot" width="320" height="220" viewBox="0 0 222 158" >
        <path id="sliderFill" stroke-width="2" stroke="grey" fill="lightgreen" d="M 0 170 A 10 20 0 0 1 0 170 L 110 110 Z"/>
    </svg>
    
    <script>
        SVG.get("sliderFill").animate().attr('d','M 0 170 A 10 20 0 0 1 '+ 70 + ' ' + 30 +' L 110 110 Z')
    </script>
</body>
</html>

RmiTtro:
The commit b0ca7491ace6f404c82607f59b22aaa3419ad0c0 should fix this issue. When I added the ability to animate path to svg.js I completely forgot that you could also use the `attr` method. :flushed:

PR #854, opened 2018-05-09
fix typings

pycraft114:


svg.js.d.ts
```diff
@@ -105,7 +105,7 @@ declare namespace svgjs {
     }
     interface Element {
         bbox(): BBox;
-        rbox(): RBox;
+        rbox(comparison?: any): RBox;
```
Fuzzyma: Instead of `any` you could use `Element` here, right?

pycraft114: hmm.. not so sure. "This will return an instance of SVG.RBox containing the following values relative either to the screen or the SVG.Container parameter" says the official documentation. So Im not sure if SVG.Container and only SVG.Container should be passed or not.

Fuzzyma: Yes ofc you are right. SVG.Container is the correct thing to pass here. I just wanted to say, that `any` might be to broad

pycraft114: I declared it to ```any``` because i wanted to tested out if the ```rbox``` method does truly work when "only" SVG.Container is provided. Passing SVG.Image as a parameter seems to be working fine. Is SVG.Container only acceptable parameter?

Fuzzyma: The box is transformed to the coordinate space of whatever you pass in the function. So yes it is possible to pass anything which is of type `SVG.Element`. The most common usecase however is SVG.Container. It might be more correct to have it as `SVG.Element` though.

So please make this parameter of type `SVG.Element`

pycraft114: Sure thing : )

Fuzzyma:
Are you plannig to add even more commits or is this ready to get merged?

pycraft114:
This is ready to get merged. for now : )

Fuzzyma:
Thank you very much!

Issue #840, opened 2018-04-01
Using reference() method when reference doesn't exsit throws TypeError: url is undefined


Zythyr:
I am using the reference() method (http://svgjs.com/referencing/#reference) to check to see if a node has a specific type of reference (ex: mask): `rect1.reference('mask')`

reference() throws a `TypeError: url is undefined` if the reference does NOT exist. If it does exist, this method works fine... 

See example below: https://codepen.io/anon/pen/YaLJNN 
```
var draw = SVG('drawing').size(400, 400) // Root SVG

var rect1 = draw.rect(100, 100).move(10, 10).fill('red')
var rect2 = draw.rect(100, 100).move(10, 150).fill('green')
circle = draw.circle(150).move(50, 50).fill('#FFF')

rect1.maskWith(circle) // apply a mask to rect1

console.clear()
console.log(rect1.reference('mask')) // rect1 DOES have a 'mask' reference
console.log(rect2.reference('mask')) // rect2 does NOT have a 'mas' reference 
```

I am unsure if this issue is a bug or intentional. I suggest the following improvement: 

- If node.reference('type') does NOT exist, then output should be `null `or `undefined `
- If node.reference() method is used without an argument, all URI references on that node should be returned

Zythyr:
@Fuzzyma Thanks for following up. I created a PR. This is my first contribution in open source code. Hopefully I didn't break anything. lol :)

Fuzzyma:
Thank you very much :). A try / catch is of course a save and easy way to catch an error. However you should avoid them and ask why this error is produced. I digged a bit and found, that the `idFromReference` which extracts the id from a reference looks like this:

```js
var m = url.toString().match(SVG.regex.reference)
```

This means when you pass an undefined url it will fail.
You can fix this issue by ensuring that url is defined. e.g.

```js
var m = (url || '').toString().match(SVG.regex.reference)
```

This way it works like a charm :). If you like you can change your PR.

Fuzzyma:
I pushed a fix to master. Thanks for your contribution!

Zythyr:
Thanks for the fix. Sorry it took me so long to get to submitting a PR.

saivan:
Yes and a big thank you from me too :)

PR #857, opened 2018-05-21
reference() method should return null if reference doesn't exsit

Zythyr:
Problem: reference() throws TypeError  if reference doesn't exist. More details: https://github.com/svgdotjs/svg.js/issues/840 

This PR fixes the issue by adding a try/catch block in the reference() method. If reference doesn't exist, it should return `null `or `undefined `.

TODO: I am leaving it up to the maintainers of SVGjs to decide if it should return null or undefined


Fuzzyma:
fixed in 5520d76e4332c1a782899679ffa8cca2285a70ce

Issue #866, opened 2018-06-09
ES5 bug related with constructor call without `new` keyword


dawiidio:
When try to use svg.js in my [lib](https://github.com/dawiidio/o-menu) with es5 modules and webpack@4.x + babel@6.2.x I found bug related with new ECMAscript 262 feature, I mean the one that says about calling constructor function without `new` keyword. Before, when you have been calling function without `new`, `this` reference the global object (window in a browser), from now in this case `this` is set to `undefined` to prevent accidental littering `window` object.

Previous functionality has been used in `svg.js` file, [here](https://github.com/svgdotjs/svg.js/blob/master/src/svg.js#L2)

```js
// The main wrapping element
var SVG = this.SVG = function(element) { // here - `this` no longer will be referenced to window
  if (SVG.supported) {
    element = new SVG.Doc(element)

    if(!SVG.parser.draw)
      SVG.prepare()

    return element
  }
}
```
My suggest is a change from `this.SVG` to `window.SVG` because when module is wrapped into `umd.js` already gets prepared `window` object from parent function.

I can handle this today, in hour or two I'll create MR with proposed changes :)

Fuzzyma:
At the very end of svg.js there is a `return SVG` which automatically binds SVG to the window object. But `this` and `window` can be different. Thats why it was like it is because it led to trouble when including plugins (yes we wanted to litter the global object because plugin relied on it). However since these restrictions are there and will not go away we need to fix that somehow.

One way is to add an UMD wrapper to ALL plugins (phew...). The other would be to wait for 3.0 to get released. Actually I would prefer the second one.

What do you think?

dawiidio:
Ok, for my lib purposes I can use version from branch on forked version until the 3.0 version will be released

saivan:
3.0 is out. It should have your problem fixed, but if not; just continue the chat here :) I'll close for now

PR #867, opened 2018-06-09
Use 'window'  instead of 'this'

dawiidio:
Pull request connected to issue https://github.com/svgdotjs/svg.js/issues/866


Fuzzyma:
Sorry for not coming back on this. Can you rebase so that we can merge it? Otherwise I just commit it myself but I would like to try to maintain authorship

PR #870, opened 2018-06-12
Throw an error if svg() is called with a non-string argument.

fchristl:
So far, if you you call, say `svg({})`, it fails with an error like "TypeError: svg.replace is not a function". This PR will help users understand the actual cause of the error if they provide arguments of wrong types.


Fuzzyma:
Thank you for your contribution. We do not do input validation in this lib. It would really bloat the size and slows things down. Instead we try to deal with wahtever the user gives us. In case of the `svg()` method it would be useful to cast the input to a string before using it.

PR #921, opened 2018-10-26
Modifiers

### LINKED ISSUE #915: Implement export modifier for SVG.Element.svg()
We would like to have export modifiers so the user gets the possibility to change the node before eporting it. This can be useful e.g. for rounding values on a node, filtering or replacing nodes.

The following examples should give you a brief overview what should be possible:

```js
// instance is the svg object and not the node itself
element.svg((instance) => {
  instance.round(2)
})
```

If you want to execute multiple modifier you can call these in your function:

```js
element.svg((instance) => {
  doThis(instance)
  doThat(instance)
})
```

If you want to replace the node with another node you can return this new node:

```js
element.svg((instance=> {
  var newInstance = new SVG.Rect()
  return newInstance 
})
```

If you want to drop this node from exporting you can return false explicitly:

```js
// drops every node
element.svg((instance=> {
  return false
})
```

**Note: Implementation has to be done on the 3.0 branch**
For all questions head over to our gitter chat: https://gitter.im/svgdotjs/svg.js

nobuti:
This PR fixes #915. It tries to implement every scenario described in the issue's description.


nobuti:
Not sure why it's failing, could anyone help me to fix it?

Fuzzyma:
Thanks for your contribution! It seems like there was a misunderstanding relating the specs for the issue.
The code should loop through all nodes under "this" (the current node) and pass each node to the modifier. If a modifier now returns false, this node should be dropped. If a new node is returned, it should be used as a replacement.

src/element.js
```diff
@@ -271,9 +271,23 @@ SVG.Element = SVG.invent({
         for (len = well.children.length; len--;) {
           this.node.appendChild(well.firstElementChild)
         }
-
       // otherwise act as a getter
       } else {
+        // expose node modifiers
+        if (typeof svg === 'function') {
+          this.each(function () {
```
Fuzzyma: Use deep-each by passing `true` as second parameter

src/element.js
```diff
@@ -271,9 +271,23 @@ SVG.Element = SVG.invent({
...
+            well = svg(this)
+
+            if (well instanceof SVG.Element) {
+              this.replace(well)
+            }
+
+            if (typeof well === 'boolean' && !well) {
+              this.remove()
+            }
```
Fuzzyma: Check for `false` expicitely and at first, then you can simplify the replace-condition

```js
            // If modifier returns false, discard node
            if (well === false) {
              this.remove()
            }
            
            // If modifier returns new node, use it
            if (well != this) {
              this.replace(well)
            }
```

nobuti: You are right!

Fuzzyma: Oh - One thing I forgot: You have to return like `return this.remove()` or make it an else-if. Otherwise both conitions can be executed.

I think I prefer the else-if in this case

nobuti: Yes, I was checking that `well` is truthy but if-else is more readable I think :+1:

Fuzzyma:
Nice! The travis error seems to come from an internal node module erroring out. Can you run the tests locally with `npm test` and doublecheck if it succeeds?

nobuti:
I've found https://github.com/zurb/foundation-sites/issues/11555, I think it's the same issue. Locally using node 9.11.2 test task is working fine. I'm going to try node 11.

nobuti:
Confirmed, I get the same error using node 11 when running `npm run build:test`

Fuzzyma:
OK. I somewhere read that you just need to update the dependencies. Will
take care of this later. Will merge asap :)

Buti <notifications@github.com> schrieb am Sa., 27. Okt. 2018, 12:57:

Fuzzyma:
Thank you very much!

PR #923, opened 2018-11-04
790-Add-support-for-new-colour-spaces

matuss1899:


saivan:
Awesome! But there are some slight issues:
- This should be added to the 3.0.0 branch (its a 3.0 feature)
- Tests would be awesome 💃 

We wanted to modify the way animations happen on fills, strokes etc... We were thinking of using the lab space by default. What are your thoughts on that? Also, giving people the ability to make Lab gradients as well would be cool too.

Fuzzyma:
@matuss1899 its a bit unfair. We changed the 3.0 branch to es6 modules and we need this feature to be merged against the 3.0 branch. Can you please change your code so that it fits the 3.0 branch? Shouldnt be much of an effort. Otherwise we will just go ahead and copy it over but that wouldnt maintain authorship :/

Issue #796, opened 2018-02-01
Allow for the creation of elements with the attribute syntax


saivan:
Currently, we can make a rectangle with

```javascript
draw.rect(30, 50)
```

But we would also like the ability to use:

```javascript
draw.rect({width: 30, height: 50, fill: "blue"})
```

This will be equivalent to calling:

```javascript
draw.rect().attr({width: 30, height: 50, fill: "blue"})
```

PR #924, opened 2018-11-08
added possibility to pass in additional attribues to element creators (#796)

Fuzzyma:
e.g.
- `canvas.rect({x:100})` or
- `canvas.rect(100, 100, {x:100})`

Please add your thoughts!

Iam not entirely satisfied with the way I need to wrap every function on every export. However, it was the cleanest way for now


saivan:
This looks like a sensible approach to me

Fuzzyma:
Well if you are fine with that we can merge it :)

PR #925, opened 2018-11-09
svg set is back (#645)

Fuzzyma:
This implements almost every functionality described in #645. However, the `collect()` method is missing.
Instead I would like to propose a similar idea with `each()`.

When the returned value of a method is not the object it was called on, we return a new List with all the return values. That would make animations for Lists much easier e.g.

```js
new List([rect, rect, rect]).animate().move(100, 100)
```

It should work right out of the Box. We just need to add some custom Runner methods to the List in that case.


saivan:
Can I have more usage examples here. I'm not totally clear about how this works. Can you do:

``` each('.square').fill('red') ``` 

for example?

File: src/types/List.js [hunk 1]
```diff
@@ -0,0 +1,38 @@
...
+  this.length = 0
+  this.push(...arr)
+})
+
+export default List
+
+extend(List, {
+  each (cbOrName, ...args) {
+    if (typeof cbOrName === 'function') {
+      this.forEach((el) => { cbOrName.call(el, el) })
```
saivan: What is cbOrName? Give me a better variable name here :P

src/types/List.js
```diff
@@ -0,0 +1,38 @@
+import { extend } from '../utils/adopter.js'
+import { subClassArray } from './ArrayPolyfill.js'
+
+const List = subClassArray('List', Array, function (arr) {
+  this.length = 0
+  this.push(...arr)
+})
+
+export default List
```
saivan: Maybe there should be tests for this?

src/types/List.js
Fuzzyma: Its a callback or a name - what do you want more? :D 
Okeee - maybe `cbOrMethodName`

src/types/List.js
Fuzzyma: Maybe there should be tests for the whole fx module ;)

Fuzzyma:
Actually the List does not function any different from the old Set. The only exception is the `each` function which can take a method name with corresponding parameters now.

Additionally you have full array functionality (more in new than in old browsers)

```js
let list = new List([a, b, c, d])

// call methods
list.fill('blue').move(0, 0)

// even create new objects in all elements from list
list.rect(100, 100)

// or use `each`
list.each(function () {
  // do something
})

list.each('fill', 'red')

// have plain array methods
list.forEach(...)
list.map(...)
list.reduce(...)
list.sort(...)
```

saivan:
Wow okay. Lets merge this! :P !

Fuzzyma:
After discussing with saivan we agreed, that its best to return a new list when a method was called. This list holds all values which were returned by the method call.

That will look like this:

```js
let a = new List([Rect, Circle])
let b = a.cx() // == List([10, 20])
let runners = a.animate() // == List([Runner, Runner])
runners.move(100, 100)
```

If the use really needs performance and dont want to generate a new list on every call, they can just loop through the List (its an array) or use forEach

PR #927, opened 2018-11-13
Add node document-fragment type check

NorbertB29:
It solves the `node.getAttribute is not a function` issue in Shadow DOM.

### DIFF
--- a/src/doc.js
+++ b/src/doc.js
@@ -53,7 +53,7 @@ SVG.Doc = SVG.invent({
     }
     // custom parent method
   , parent: function() {
-      if(!this.node.parentNode || this.node.parentNode.nodeName == '#document') return null
+      if(!this.node.parentNode || this.node.parentNode.nodeName == '#document' || this.node.parentNode.nodeName == '#document-fragment') return null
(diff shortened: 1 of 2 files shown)


NorbertB29:
> This sounds useful. Can you give me one concrete piece of sample code?

Yes, we are using svg.js in polymer 2 and we successfully reproduced this issue in this jsfiddle:
[https://jsfiddle.net/vtqyg90m/1/](url)

Fuzzyma:
Mh when i load this page I just get an "Polymer is undefined" error

NorbertB29:
> Mh when i load this page I just get an "Polymer is undefined" error

Try in Chrome please. It works for me.

Fuzzyma:
Ok, forgot to ask. Can you add tests?

NorbertB29:
I opened a new PR for the unit tests please check it. Also we would like ask you to make a patch release with this fix as soon as possible please. Thank you!

Fuzzyma:
[Here you go](https://github.com/svgdotjs/svg.js/releases/tag/2.7.0)

PR #945, opened 2018-12-13
Add PointArray.transform by analogy to Point.transform

edemaine:
I recently had occasion to want to transform the points in an `SVG.Polygon` according to a transformation matrix.  (The motivation is to "inline" `transform` attributes into the point coordinates themselves, for software that supports SVG but not the `transform` attribute.)  `SVG.Polygon.array` reveals the `SVG.PointArray` that represents the coordinates, and `SVG.Polygon.plot` lets me put a transformed `PointArray` back into the polygon.  What's missing is an `SVG.PointArray.transform` (essentially a loop over `SVG.Point.transform`) to transform the `PointArray` according to a given matrix.

This PR adds this functionality.  If you like the proposed addition in principle, I can write tests and documentation.

### DIFF
--- a/src/types/PointArray.js
+++ b/src/types/PointArray.js
@@ -73,12 +73,14 @@ extend(PointArray, {
 
   // transform points with matrix (similar to Point.transform)
   transform (m) {
-    let points = []
+    const points = []
 
     for (let point of this) {
       // Perform the matrix multiplication
-      points.push([m.a * point.x + m.c * point.y + m.e,
-        m.b * point.x + m.d * point.y + m.f])
+      points.push([
+        m.a * point[0] + m.c * point[1] + m.e,
+        m.b * point[0] + m.d * point[1] + m.f
+      ])
     }
 
     // Return the required point
--- a/src/types/PointArray.js
+++ b/src/types/PointArray.js
@@ -78,12 +78,12 @@ extend(PointArray, {
     for (let point of this) {
       // Perform the matrix multiplication
       points.push([m.a * point.x + m.c * point.y + m.e,
-                   m.b * point.x + m.d * point.y + m.f])
+        m.b * point.x + m.d * point.y + m.f])
     }
 
     // Return the required point
     return new PointArray(points)
-  }
+  },
 
   // Move point string
   move (x, y) {
--- a/src/types/PointArray.js
+++ b/src/types/PointArray.js
@@ -75,7 +75,8 @@ extend(PointArray, {
   transform (m) {
     const points = []
 
-    for (let point of this) {
+    for (let i = 0; i < this.length; i++) {
+      const point = this[i]
       // Perform the matrix multiplication
       points.push([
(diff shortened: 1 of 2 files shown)


src/types/PointArray.js
```diff
@@ -71,6 +71,20 @@ extend(PointArray, {
...
+
+    for (let point of this) {
+      // Perform the matrix multiplication
+      points.push([m.a * point.x + m.c * point.y + m.e,
+                   m.b * point.x + m.d * point.y + m.f])
+    }
+
+    // Return the required point
+    return new PointArray(points)
+  }
```
Fuzzyma: comma needed!

edemaine:
Fixed the missing comma, as well as a critical bug encountered while testing.  Added a few simple tests.

Transforming a path would be a cool feature too, but I think it's not nearly as easy.  In particular, transforming an arc (`a`/`A`) command according to an arbitrary matrix would be quite challenging -- perhaps not even possible?  (I'm not sure whether an elliptical arc could transform to an hyperbolic arc...)

Anyway, it's inconsistent to have just `SVG.Point.transform`, so this at least corrects one such inconsistency. 😄 Maybe a future PR can target `SVG.PathArray.transform`.

Fuzzyma:
I actually once created such a feature for paths. You can take a look at svg.screenbbox.js. As you already mentioned, it seems to be hard to transform arc commands. But then I figured, that I only transform the start and end points of arcs. Yes, the arc might not be correct because you cant really draw skews arcs, but it was better than nothing^^.

To your implementation: I just saw, you used for...of loops. While this is not really a problem, it is anyway.
PointArray is an Array in modern browsers but its not in old Browsers. In old Browsers its an array-like object. As hard as it is, please use a normal for loop to make sure, it works in old Browsers, too.

For your inconsisteny argument: I dont really agree. But thats just my opinion^^. Its a useful feature, so why not...

edemaine:
Hmm, I also just realized that all the other similar methods of `PointArray`, like `PointArray.move`, are in-place modification operations (while also returning `this`), so `PointArray.transform` should probably also modify in-place?

On the other hand, `Point.transform` makes a copy of the point (which is why I made a copy transform), so I'm not totally sure which is best.  Any preference?

Fuzzyma:
No problem, it **IS** really a hack. If you take a look into ArrayPolyfill.js you will see how ugly that hack is. I even use the Function constructor. But I found it worth the effort because having native arrays where it is supported is awesome!

I would prefer to have it as a seperate PR. This one is fine already :)

Fuzzyma:
> Hmm, I also just realized that all the other similar methods of PointArray, like PointArray.move, are in-place modification operations (while also returning this), so PointArray.transform should probably also modify in-place?

Good question... I dont really have a preference because of your mentioned inconsistency. However, Matrix also returns a new one on transformation. So lets stay consistent here.
I see there is much to fix in v4.0 lol

Issue #949, opened 2019-01-06
How to get new version? I see only 2.7.1 in my package.json


PiotrFr:
I can't put version 3.0.5 in to my package.json. I get information that only 2.7.1 I can choose

![image](https://user-images.githubusercontent.com/13883511/50735290-effea180-11ac-11e9-9e5b-dcc8b83cbf6b.png)

chenz24:
I installed v3.05.  but import SVG from '@svgdotjs/svg.js';  got error.  when I use v2.7 , it work fine.
when I switched to v3, import did not work properly. I don't know why

Fuzzyma:
Without code I can't help much :/

Fuzzyma:
Fixed with latest release

PR #956, opened 2019-02-08
use umd format in rollup. Update .d.ts file with changes for 3.0

kabentley:
I attempted to use 3.0.11 with TypeScript. I think the rollup.conf.js file has the wrong format after your changes to use esm modules. Using 'umd' allows it to be imported correctly.

I also updated some of the obvious changes to the .d.ts from the changelog. I have to confess my changes are not complete, but they allowed me to use 3.0.11 in my application.

### DIFF
--- a/.config/rollup.config.js
+++ b/.config/rollup.config.js
@@ -76,7 +76,7 @@ const config = (node, min) => ({
     file: node ? './dist/svg.node.js'
       : min ? './dist/svg.min.js'
         : './dist/svg.js',
-    format: node ? 'cjs' : 'iife',
+    format: node ? 'cjs' : 'umd',
     name: 'SVG',
     sourcemap: true,
     banner: headerLong,
--- a/.config/rollup.config.js
+++ b/.config/rollup.config.js
@@ -76,7 +76,7 @@ const config = (node, min) => ({
     file: node ? './dist/svg.node.js'
       : min ? './dist/svg.min.js'
         : './dist/svg.js',
-    format: node ? 'cjs' : 'umd',
+    format: node ? 'cjs' : 'iife',
     name: 'SVG',
     sourcemap: true,
     banner: headerLong,
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   "main": "dist/svg.node.js",
   "unpkg": "dist/svg.min.js",
   "jsdelivr": "dist/svg.min.js",
-  "browser": "dist/svg.min.js",
+  "browser": "src/main.js",
   "module": "src/main.js",
   "maintainers": [
     {
--- a/svg.js.d.ts
+++ b/svg.js.d.ts
@@ -1,13 +1,14 @@
 export = svgjs;
 export as namespace svgjs;
 
-declare var svgjs: svgjs.Library;
+declare const svgjs: svgjs.Library;
 
 // todo add SVG.FX
 declare namespace svgjs {
     export interface Library {
-        (id: string): Doc;
-        (domElement: HTMLElement): Doc;
+        (id: string): Svg;
+        (domElement: HTMLElement): Svg;
+        (): Svg;
         ns: string;
         xmlns: string;
         xlink: string;
@@ -37,16 +38,17 @@ declare namespace svgjs {
         siblings(): Element[];
         position(): number;
         next(): Element;
-        previous(): Element;
+        prev(): Element;
         before(element: Element): Element;
         after(element: Element): Element;
+        defs(): Defs;
     }
 
     // array.js
     type ArrayAlias = _Array | number[] | string;
 
     interface _Array {
-        new (array?: ArrayAlias, fallback?: number[]): _Array;
+        new(array?: ArrayAlias, fallback?: number[]): _Array;
         value: number[];
         morph(array: number[]): this;
         settle(): number[];
@@ -54,7 +56,6 @@ declare namespace svgjs {
         toString(): string;
         valueOf(): number[];
         parse(array: ArrayAlias): number[];
-        split(string: string): number[];
         reverse(): this;
         clone(): _Array;
     }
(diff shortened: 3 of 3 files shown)


Fuzzyma:
Thanks for fixing the d.ts file. I am not sure about the umd wrapper though. The library is not meant to be consumed by typescript as a bundle. Actually you should be able to import the lib as esm import. Unfortunately iam not good with typescript so I dont really know how this is all supposed to work

kabentley:
The umd change has nothing to do with TypeScript. You changed the library to be a module, but your rollup config says its an iife. From what I can tell, without this change your library can't be imported from JavaScript and can only be used with a Script tag. I actually think you should use umd format for Node too.

I'm pretty sure this is the problem with issue #949

Fuzzyma:
No, if you take a look into the package.json you can see that there a a few keys specifying which file to load for what. e.g. node loads the whats specified with the `main` key which is svg.node.js taht has a module export. The `module` key specifies `src/main.js` which is the esm import. As far as I know that should also be the one imported by typescript. And then you have the normal browser bundle which is ofc only an iife because everything else would make no sense.

kabentley:
your package.json file also has:

  `"browser": "dist/svg.min.js",`

which tells webpack to use the minified, rolled-up version. That's the version that gets imported for anyone who webpacks (again, this has nothing to do with TypeScript).

You should use "umd" format for your rollup config for both Node and browsers.

Fuzzyma:
> You should use "umd" format for your rollup config for both Node and browsers.

We didnt create multiple bundles to use umd in the end. Umd is used to make packages work in node. Thats not needed for our bundle. The bundle is solely there to be loaded with a script tag. Every module bundler and even typescript should use /src/main.js. If webpack uses the wrong key, tell webpack to load the correct instead.
I dont know what typescript is doing but it should load /src/main.js, too

kabentley:
OK, I officially give up trying to convince you to use umd (you seem to think there's virtue in iife over umd, i can't see why).

Regardless, if you really don't want bundlers to use your rollup file, then you should remove your "browser" field in your package.json file because that's its sole purpose (see [here](https://github.com/defunctzombie/package-browser-field-spec)).

And please stop conflating this issue with TypeScript. TypeScript has nothing to do with bundling/loading. It merely creates JavaScript.

Fuzzyma:
Dont get me wrong. I am very happy about your attemps to get 3.x running with typescript.
Obviously I am missinformed about the browser keyword and will remove it. However, your liked site states, that browser replaces the `main` key which on the other hand is also the wrong file (because it is the node file). Maybe I should put src/main.js into browser?

I am aware, that typescript is compiled to javascript. However, I am not aware about the way the typescript compiler resolves packages. Does it even resolves packages or does it compiled into native esm imports? And if so, how does it know where to find e.g. `@svgdotjs/svgjs`? You see, I dont have the knowledge to really solve that.

For the umd thing I have a strong opinion about that. There is no benefit to have an umd wrapper in the browser version

kabentley:
When you compile/transpile your .ts into .js, you tell the compiler the module format you want to target. If you specify "commonJs" it generates requires, if you specify "es6" it generates import statements. In that case the generated .js will just have "import * as SVG from "@svgdotjs/svg.js", etc. That's why I keep saying this is not a TypeScript issue. All consumers of your library who try to import/require your bundle will have the same problem. If you changed to "umd" format it could be used either as a Script tag or imported. That's the benefit of using the (now recommended) umd format.

Anyway, I think you're right it is better to import it as an esm anyway. I changed my PR to have:
`  "browser": "src/main.js",`

reverted the change from iife to umd, and then changed the d.ts file to be for the esm module. 

The only difference is that if you do `import * as SVG from "@svgdotjs/svg.js"` you now have to call SVG.SVG().

Fuzzyma:
Yes thats the issue with the main.js. Its meant to import everything without default export. We did that so that to enable treeshaking for bundlers. However it seems like everyone prefers the big SVG object over direct imports. What do you think about that?

I mean, we have the src/svg.js file which basically exports that big object. However, it denies any use of treeshaking

kabentley:
I think what you have now is much better and much more modern than one big function object (there's still one big object, it's just called "exports".) Default exports are not helpful.

You do need to update the documentation for the 3.0 changes. But the switch from `SVG()` to `new SVG.Svg()` seems minor compared to some of the other changes you made.

BTW, my changes to the .d.ts file were made in haste. Someone who really knows the library (I've only used a very small slice) should review them carefully. And really, TypeScript is the greatest thing to come to JavaScript ever, IMHO.

Fuzzyma:
You dont even need `new SVG.Svg()`. You can still use `SVG.SVG()` which is the old SVG function. `SVG.Svg()` is the constructor for a new `<svg>` node. It works quite differently.

So if people really want there old object bag, they could simply go `const lib = SVG.SVG` and `Object.assign(lib, SVG)`. Thats basically what i made in `src/svg.js` to create the standalone bundle.

I am aware, that the d.ts file is full of errors. It never was complete. So your changes only did improve the situation - no worries about that :D.

I thought about adding module-wise typings for svg.js but thats not easily possible because of the way the lib is composed. Modules sometimes add functionality to other modules and so on. I have no clue how to describe this in a d.ts file

kabentley:
Writing type declarations is a lot like writing documentation. In fact, if you put the documentation is in the type file, tools like Visual Studio Code show it while you're editing.

Anything inside a `declare` statement does not generate any code, only type information. You can have functions, variables, type aliases, interfaces and classes. If you have both a class and an interface with the same name they are combined. Whomever created the original .d.ts. did that a lot, but i prefer to put everything in the class declaration, since that makes it easier to understand when you're looking at (you'll see i did that for the Element class but no others).

Modules extending other modules isn't tricky, just declare it however you'd write the documentation for it.

Any chance you'll accept this PR now?

Fuzzyma:
Sure, but you keep pushing updates so I didnt know when to merge :D

kabentley:
ok, I'm done!

PR #957, opened 2019-02-12
cleanup of d.ts

kabentley:
Sorry, I found more errors in the d.ts file when I tried to use it.

### DIFF
--- a/svg.js.d.ts
+++ b/svg.js.d.ts
@@ -87,7 +87,7 @@ declare module "@svgdotjs/svg.js" {
     type ArrayAlias = _Array | number[] | string;
 
     class _Array {
-        constructor(array?: ArrayAlias, fallback?: number[]): _Array;
+        constructor(array?: ArrayAlias, fallback?: number[]);
         value: number[];
         morph(array: number[]): this;
         settle(): number[];
@@ -100,9 +100,9 @@ declare module "@svgdotjs/svg.js" {
     }
 
     class Dom {
-        constructor(element: string, inherit?: any): Dom;
+        constructor(element: string, inherit?: any);
         words(text: string): this;
-        element(element: string, inherit?: Object): Dom;
+        element(element: string, inherit?: Object): this;
         addTo(parent: Dom): this;
         putIn(parent: Dom): Dom;
         children(): Element[];
@@ -119,7 +119,7 @@ declare module "@svgdotjs/svg.js" {
     }
 
     // boxes.js
-    interface Box {
+    class Box {
         height: number;
         width: number;
         y: number;
@@ -132,21 +132,15 @@ declare module "@svgdotjs/svg.js" {
         y2: number;
         merge(box: Box): Box;
         transform(m: Matrix): Box
-    }
-
-    class BBox extends Box {
-        constructor(element?: Element): BBox;
-    }
-    class RBox extends Box {
-        constructor(element?: Element): RBox;
-    }
-    class TBox extends Box {
-        constructor(element?: Element): TBox;
+        Box(source: string);
+        Box(source: []);
+        Box(source: object);
+        Box(x: number, y: number, width: number, height: number);
     }
 
     // clip.js
     class ClipPath extends Container {
-        constructor(): ClipPath;
+        constructor();
         targets: Element[];
         remove(): this;
     }
@@ -163,9 +157,12 @@ declare module "@svgdotjs/svg.js" {
 
     type ColorAlias = string | ColorLike;
 
-    class Color extends ColorLike {
-        constructor(): Color;
-        constructor(color: ColorAlias): Color;
+    class Color implements ColorLike {
+        r: number;
+        g: number;
+        b: number;
+        constructor();
+        constructor(color: ColorAlias);
 
         toString(): string;
         toHex(): string;
@@ -188,9 +185,9 @@ declare module "@svgdotjs/svg.js" {
     class Defs extends Container { }
 
     class Svg extends Container {
-        constructor(): Svg;
-        constructor(id: string): Svg;
-        constructor(domElement: HTMLElement): Doc;
+        constructor();
+        constructor(id: string);
+        constructor(domElement: HTMLElement);
         namespace(): this;
         defs(): Defs;
         parent(): HTMLElement;
@@ -215,7 +212,7 @@ declare module "@svgdotjs/svg.js" {
         attr(obj: Object[]): Object;
         back(): this;
         backward(): this;
-        bbox(): BBox;
+        bbox(): Box;
         before(element: Element): Element;
         center(x: number, y: number): this;
         classes(): string[];
@@ -292,7 +289,7 @@ declare module "@svgdotjs/svg.js" {
         point(x: number, y: number): Point;
         position(): number;
         prev(): Element;
-        rbox(element?: Element): RBox;
+        rbox(element?: Element): Box;
         reference(type: string): Element;
         remember(name: string, value: any): this;
         remember(name: string): any;
@@ -315,7 +312,7 @@ declare module "@svgdotjs/svg.js" {
         stroke(stroke: StrokeData): this;
         svg(): string;
         svg(svg: string): this;
-        tbox(): TBox;
+        tbox(): Box;
         toggleClass(name: string): this;
         toParent(parent: Dom): this;
         toString(): string;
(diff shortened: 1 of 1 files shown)


Fuzzyma:
Just tell me when you are ready to merge ;)

kabentley:
ok, I'll let you know. ;)

kabentley:
Well there's still a lot of cleanup to be done, but I think it's better now than before. Unfortunately I'm just not going to be able to spend more time on it now. 

BTW, I didn't say it before but this is a very nice library. THANK YOU all for your efforts!

kabentley:
Will you be making a new release any time soon?

Fuzzyma:
Thank you very much - for your words and for your contribution :). I think the fact, that 3.x is finally usable with typescript is worth a release :D. However I will wait 2 more days just in case you cannot stop yourself from fixing more :P

cmidt-veasna:
@kabentley Correct me if I'm wrong, I believe the previous declaration with **namespace** is the right way to declare types definitions file according to this [template](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html) and according to this [template](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-plugin-d-ts.html) `declare module 'someModule'` is suppose to be use as add-on or plugin.

As my understanding, the syntax `declare module 'someModule'` is used for a few different use case:
1. the library itself does not provide any support on typescript so we need module plugin to enable strong type check and avoid the use of **any**.
2. we want to extend or prototype existing exported class or module where the function or properties did not existing in first place in the module library. For example: SVG.JS provide a feature to `extend` or `invent` on existing class thus the provided typescript definition won't be able to cover the new properties or function that we create with either `extend` or `invent`.

cmidt-veasna:
@kabentley @Fuzzyma Current I'm working on including typescript test by utilize **webpack** compiled. The test was working fine with previous syntax however the import module does not work with module plugin syntax perhaps that because the module plugin syntax suppose to be only use and declare outside of module or library itself.

kabentley:
@cmidt-veasna having a test suite for the .d.ts file would be wonderful. Do you have a fork where i can look at your test? 

You cannot import the files from the "dist" directory (see long discussion  here #956). You must import from the src/main.js module, which is what the `declare module` describes. For webpack, the relevant change was the `browser:` keyword in the package.json file.

cmidt-veasna:
@kabentley Sorry for taking a bit too long as I try to fix thing and merge with current update. The compile of types is not working with current module plugin style, it keep raising an error `"@svgdotjs/svg.js" is not a module`. If you would like to take a look for test code please visit [here](https://github.com/cmidt-veasna/svg.js/tree/typescript-test).

I add a script command `test:types` with ChromeHeadless and FirefoxHeadless only. The current test is taking from Javascript test, it almost identical to the javascript test. 

As for the definition is almost align with Javascript function however there will be a few thing to add and correct. Let me know if you found solution or able to make it work module plugin style. Also notice that my update was take the same style as previously as this is my original update before you have push a new update to the repo.

kabentley:
@cmidt-veasna somehow your alias in your .config file isn't working. It should resolve to  '../../src/main.js`, not `../../src/svg.js`

You can try changing:

`import * as SVGJS from '@svgdotjs/svg.js'`
to:
`import * as SVGJS from '../../src/main.js`

kabentley:
@cmidt-veasna But wait, you don't have my changes to the svg.js.d.ts file?

Fuzzyma:
@kabentley I didn't release a new version yet so he could still use the old one

kabentley:
actually, I just pulled your repo, and copied my d.ts into it and it works fine except it shows all of the 3.0 changes that your tests don't reflect.

cmidt-veasna:
@kabentley actually there are many error occurred during the merge as I've been working on this before your merge happen. So technically your change was still there however there a few thing I revert back to use interface not class. @kabentley you're right, apparently, the module plugin style seem to be working with current config. It seem that typescript compiler may have different behavior when lookup for the dependencies. So the below is the sequential of how the `.d.ts` file up end like it is right now.

1. First, originally the import module is using relative import `../..`. The idea behind this import is that svg.js is a full javascript library and only expose typescript definition to allow the use in typescript project. The typescript compiler is working similar way to NodeJS to resolve module or dependencies, the important part of it is the file `package.json` as the compiler is looking for. So I intended to make the test work the same way as `svg.js` being imported as module on third party project and with that it 's ensure that the configure in `package.json` for types, module or webpack as well as the definition file itself should as expected.
2. After I sync with the upstream then I realize `.d.ts` has been change, so I tried to merge and keep it as you did however typescript keep raising the error `svg.js.d.ts` is not a module. so I tried to find solution for it but not really there so I decide to revert the declaration into namespace similar to previous one then other error occurred, with the namespace define the class definition is not working with it only interface seem to working fine. That's was the reason I change all class definition into interface instead.
3. The test seem to work fine but I don't really like the relative import `../..` so update `.tsconfig` to trick the typescript compiler to find `package.json` when import as `@svgdotjs/svg.js`.

That how it's end up as right now and @kabentley somehow the trick is also resolve the problem with `svg.js.d.ts` is a not module as mention above. Perhaps typescript may have thing behind all this trick.

cmidt-veasna:
@kabentley regarding the test is not reflecting 3.0, I'm not really sure, I assume that the javascript test code is up to date for 3.0 so I mirror almost all the test of typescript to javascript test. @Fuzzyma does the javascript test is up to date ?

kabentley:
@cmidt-veasna all I did was copy the latest .d.ts file into your repository (and remove the `declare` in svg.spec.ts)  and run `npm run test:types`. All of the problems are simply incorrect type declarations in the test or incompleteness of the .d.ts file (which is what I was hoping you'd fix).

Do you want me to try to fix the tests?

cmidt-veasna:
@kabentley I haven't run the test with latest `.d.ts` just the VS Code does not hint the error anymore. However  the declare in `svg.spec.ts` should not be removed, it's a plugin style to enable `extend` feature of `svg.js`. As you can see in the file with below code:
```javascript
SVGJS.extend(SVGJS.Rect, {
     soft: function () {
          return (this as SVGJS.Element).opacity(0.2)
     }
})
```
without declare additional plugin, typescript will raise an error as function `soft` is not declare anywhere and then the only solution is to cast `Rect` object as `any` type which we tend to avoid in typescript implementation.

@kabentley  I will try with the latest `.d.ts` however I will wait @Fuzzyma to confirm about javascript test. I did look into javascript source code which seem to be align with the test code.

Fuzzyma:
Good morning :).
The tests are run against the the 3. bundle so yes, these tests are for 3.x. However, they are not written in esm style. They assume a global SVG object as in 2.x to run (thats it, because these tests are basically ports from 2.x to 3.x).
However, new tests are always added in esm style and written in es6. You can recognize them by looking at the folders in the spec directory. So all files in the spec directory are es5 tests while all folders contain es6 tests.
To run these es6 tests you can use `npm run test:es6` or you can transpile the es6 tests and run them along the es5 tests, too (npm run build:tests). OR you can run these tests by svgdom: `npm i run test:svgdom`.

Jeaah - we need to cleanup the mess. I am painly aware of that

kabentley:
@cmidt-veasna I have now combined all of the (very good!) work you did on the .d.ts file with mine in #960 . 

I had to make several changes to your tests to get them to compile, but I don't know how to get my changes into your repo.

Fuzzyma:
Awesome how you all collaborate with eachother. That's so great to see! 

You can make a PR against his repository so he can pull your changes into his repo (if this is what you want)

kabentley:
I'm a GitHub novice. I can't see how to make a PR against his repository since I can't fork it - it tells me that I've already forked it from the main SVG.js repo.

I'm making a branch in my repo so he can see what i did.

Fuzzyma:
Mh - iam not at my desktop atm so I can't test it but can you check if you can go to his repo and to his PRs and see if there is a button "create new pull request" where you can choose your branch to pull it against his repo?

kabentley:
You can see what i did [here](https://github.com/kabentley/svg.js/tree/typescript-tests).

cmidt-veasna:
@kabentley Thanks for your update. I've merge your changed. I will work on adding more type such as style and attribute before request merge with main repository. It will probably take me 1 or 2 days to read through all document from Standard SVG. After this done the constructor should be aligned with Javascript constructor.

Fuzzyma:
@kabentley @cmidt-veasna You both have to make sure, that you preserve authorship from eachothers code. It seems like both you included code from the other. Maybe you should collaborate more closely

PR #960, opened 2019-02-14
more cleanup of .d.ts

kabentley:
I found a few more ;)

### DIFF
--- a/svg.js.d.ts
+++ b/svg.js.d.ts
@@ -210,6 +210,7 @@ declare module "@svgdotjs/svg.js" {
         link(url: string): A;
         marker(width?: number, height?: number, block?: (marker: Marker) => void): Marker
         mask(): Mask;
+        nested(): Svg;
         path(): Path;
         path(d: PathArrayAlias): Path;
         pattern(width?: number, height?: number, block?: (pattern: Pattern) => void): Pattern
@@ -237,7 +238,6 @@ declare module "@svgdotjs/svg.js" {
         namespace(): this;
         defs(): Defs;
         parent(): HTMLElement;
-        spof(): this;
         remove(): this;
         isRoot(): boolean;
     }
--- a/svg.js.d.ts
+++ b/svg.js.d.ts
@@ -127,6 +127,7 @@ declare module "@svgdotjs/svg.js" {
         removeElement(element: Element): this;
         replace(element: Dom): Dom;
         svg(): string;
+        svg(svg: string): this;
         toString(): string;
         words(text: string): this;
         writeDataToDom(): this;


Fuzzyma:
I knew it :D. Should I merge right away?

kabentley:
why don't you just leave this PR open until you're ready to make a new release. I'm sure I'll keep finding more, but it's not critical that the .d.ts being 100% accurate (you can just do an "as any" where things are missing/wrong).

Fuzzyma:
Thats what I had in mind from the start - but you wanted it to be merged :P 
But no worries. I will just merge it before I release the next version. Until then I will leave it open :)

svg.js.d.ts
```diff
@@ -1,8 +1,12 @@
+declare class BuiltInArray<T> extends Array<T> { }
+
 declare module "@svgdotjs/svg.js" {
 
     function SVG(): Svg;
-    function SVG(id: string): Svg;
-    function SVG(domElement: HTMLElement): Svg;
+    function SVG(id: string): Element;
```
Fuzzyma: I would rename it from `id` to `selector`

svg.js.d.ts
```diff
@@ -82,58 +105,505 @@ declare module "@svgdotjs/svg.js" {
...
+        x: number;
+        y: number;
+        constructor();
+        constructor(position: CoordinateXY);
+        constructor(point: Point);
+        constructor(position: { x: number, y: number });
+        constructor(x: number, y?: number);
+        clone(): Point;
+        transform(matrix: Matrix): this;
+        transform0(matrix: Matrix): this;
```
Fuzzyma: Its `transformO` not `transform0` - Its an o not a zero. Its short for "optimized"

svg.js.d.ts
```diff
@@ -82,58 +105,505 @@ declare module "@svgdotjs/svg.js" {
...
+
+    type PathArrayAlias = PathArray | PathCommand[] | (string | number)[] | string;
+
+    class PathArray extends Array<PathCommand> {
+        constructor();
+        constructor(d: ArrayAlias<PathCommand> | PathArrayAlias);
+
+        // TODO: maybe typescript does not need to have such as direct access to this variable.
+        // SVG.JS required direct modify the value of a path array for performance
+        value: PathCommand[]
```
Fuzzyma: As far as I know this attribute doesnt exist anymore. If is does, its a bug!

svg.js.d.ts
```diff
@@ -82,58 +105,505 @@ declare module "@svgdotjs/svg.js" {
...
+        d?: number;
+        e?: number;
+        f?: number;
+    }
+
+    interface MatrixExtract extends TransformData, MatrixLike { }
+
+    type FlipType = 'both' | 'x' | 'y' | boolean
+    type ArrayXY = [number, number]
+    type CoordinateXY = ArrayXY | { x: number, y: number }
```
Fuzzyma: You used that type at the Point interface but also had one constructor extra with only `{ x: number, y: number }`. Couldnt you get rid of that extra constructor?

svg.js.d.ts
```diff
@@ -82,58 +105,505 @@ declare module "@svgdotjs/svg.js" {
...
+        valueOf(): MatrixLike;
+        to(a: any): Morphable;
     }
 
-    // boxes.js
-    class Box {
+    // List.js
+    interface List<T> extends Array<T> {
+        // TODO: check whether each is necessary to expose if it useful to external usage.
+        // each(fn: Function, ...args: any[]): void
```
Fuzzyma: Yes, each is totally usable. Its for the case, that you didnt updated the methods in the List prototype and therefore go the way over each. E.g. this will do the same:

```js
list.fill('red')
list.each('fill', 'red')
```

It can also be used to run a custom function over the whole list like map does:

```js
list.each(fn)
```

Ofc you could use map directly in that case... anyway - I will put it in the docs and so it will be public api

svg.js.d.ts
```diff
@@ -144,46 +614,421 @@ declare module "@svgdotjs/svg.js" {
...
+        constructor(source: DOMRect);
         constructor(x: number, y: number, width: number, height: number);
+
+        merge(box: Box): Box;
+        transform(m: Matrix): Box
         addOffset(): this;
         toString(): string;
         toArray(): number[];
         isNulled(): boolean;
+        to(v: MorphValueLike): Morphable;
```
Fuzzyma: Actually you can only morph a Box to a boxlike.
The other `to` functions are all with any but as a rule of thumb you can say: Everything what goes for the constructor also goes for the `to` method of that class

svg.js.d.ts
```diff
@@ -144,46 +614,421 @@ declare module "@svgdotjs/svg.js" {
...
+        constructor()
+        constructor(a: number[])
+        constructor(a: TransformData)
+
+        defaults: TransformData
+
+        toArray(): number[]
+
+        to(t: TransformData): Morphable
+        fromArray(t: TransformData): this
```
Fuzzyma: fromArray takes an array as input. Beside that, this looks fine

svg.js.d.ts
```diff
@@ -144,46 +614,421 @@ declare module "@svgdotjs/svg.js" {
...
+        clearTimeout(o: Object): void
+        cancelImmediate(o: Object): void
+    }
+
+    type ParentTypeAlias = Svg | G | A;
+
+    /**
+     * Just fancy type alias to refer to css query selector.
+     */
+    type QuerySelector = string
```
Fuzzyma: Jup, use that one for `SVG()`, too :)

svg.js.d.ts
```diff
@@ -144,46 +614,421 @@ declare module "@svgdotjs/svg.js" {
...
+        putIn<T extends this>(parent: PutInMapping<T>): PutInResultMapping<T>;
+
+        remove(): this;
+        removeElement(element: Element): this;
+        replace<T extends Dom>(element: T): T;
+        round(precision?: number, map?: string[]): this
+        svg(): string;
+        svg(a: boolean, outer?: boolean): string
+        svg(a: null | Function, outer?: boolean): string
+        svg(a: string, outer?: boolean): string
```
Fuzzyma: This is an import not an export. It returns the element itself (or in case outer=true it would return the parent)

svg.js.d.ts
```diff
@@ -144,46 +614,421 @@ declare module "@svgdotjs/svg.js" {
...
-        at(pos: number): Color;
+        words(text: string): this;
+        writeDataToDom(): this;
+
+        // prototype extend Attribute in attr.js
+        /**
+         * Get the attribute object of SVG Element. The return object will be vary based on
+         * the instance itself. For example, G element will only return GlobalAttr where Rect
+         * will return RectAttr instead.
+         */
```
Fuzzyma: This seems like this becomes super complicated. What about custom attributes and data-attributes? Can you cover all this? I mean - I dont mind as long as this works :)

svg.js.d.ts
```diff
@@ -583,72 +1421,62 @@ declare module "@svgdotjs/svg.js" {
...
-        size(width: number, height: number): this;
+        size(width: number, height?: number): this;
         marker(position: string, width?: number, height?: number, block?: (marker: Marker) => void): Marker;
         marker(position: string, marker: Marker): Marker;
     }
     // rect.js
     class Rect extends Shape {
-        radius(x: number, y?: number): this;
-    }
+        constructor(val: { width: number, height: number, x: number, y: number })
```
Fuzzyma: I am not sure about this one. Every constructor can take an object which is passed directly to `attr`. So this is not limmited to width, height...

Alternatively you can pass in a node into the Rect constructor.

This both statements are true for all elements in svg.js. So you can do the following:

```js
new Rect(SVGRectNode)
new Rect({width: 100, fill: 'none'})
```

svg.js.d.ts
kabentley: good point :)

svg.js.d.ts
kabentley: OK, I was confused about this and just trying to get the test to compile. I'll fix it

svg.js.d.ts
kabentley: would that be an array of numbers?

Fuzzyma: yes!

svg.js.d.ts
kabentley: should it return an Svg?

svg.js.d.ts
kabentley: I think he gave up and just returned "any"

svg.js.d.ts
Fuzzyma: It returns `this` for outer=false and `Element` for outer=true. I am not sure if you can type that with Typescript.
But it will always return an `Element`. Btw: If you want we can also chat on gitter :)

svg.js.d.ts
Fuzzyma: Well thats fine for now I guess :D

svg.js.d.ts
kabentley: gitter sounds like a better choice. How do I do that?

kabentley: I added this logic

Fuzzyma: You can find our gitter channel here: https://gitter.im/svgdotjs/svg.js

You can login with github :)

svg.js.d.ts
```diff
@@ -668,13 +668,10 @@ declare module "@svgdotjs/svg.js" {
...
         constructor(a: number[])
         constructor(a: TransformData)
-
         defaults: TransformData
-
         toArray(): number[]
-
         to(t: TransformData): Morphable
-        fromArray(t: TransformData): this
+        fromArray(t: number[] | Object ): this
```
Fuzzyma: Why do you have this | Object there? Shouldnt only teh number array be enoough?

svg.js.d.ts
```diff
@@ -1455,22 +1453,23 @@ declare module "@svgdotjs/svg.js" {
...
         move(x: number, y: number): this;
         size(width: number, height?: number): this;
         marker(position: string, width?: number, height?: number, block?: (marker: Marker) => void): Marker;
         marker(position: string, marker: Marker): Marker;
     }
-    // rect.js
+
     class Rect extends Shape {
-        constructor(val: { width: number, height: number, x: number, y: number })
+        constructor(val?: SVGRectElement);
```
Fuzzyma: As already mentioned, this possibility exists for all elements :)

svg.js.d.ts
```diff
@@ -1243,6 +1255,7 @@ declare module "@svgdotjs/svg.js" {
         stop(jumpToEnd: boolean, clearQueue: boolean): Animation;
         stop(offset?: NumberAlias | string, color?: NumberAlias, opacity?: NumberAlias): Stop;
         stop(val: { offset?: NumberAlias | string, color?: NumberAlias, opacity?: NumberAlias }): Stop;
+        stroke(): any;
         stroke(color: string): this;
         stroke(stroke: StrokeData): this;
         tbox(): Box;
```
Fuzzyma: This one can go. It doesnt exist anymore

kabentley: I can take it out, but it looks to me like it is there. Or do i misunderstand?

https://github.com/svgdotjs/svg.js/blob/9e75004466f251d765f72fd80fad846e71fae107/src/modules/optional/sugar.js#L24

Fuzzyma: You have a reference to sugar.js in your answer - is this wanted? Because I dont get it :D.
We dont have a tbox. We have rbox and bbox. tbox was deprectated long time ago and replaced by rbox

kabentley: i'm sorry, i thought you were referring to the green line (`stroke(): any`). Got it :)

Fuzzyma: Ah okay :D. Btw, can we meet at gitter for a few minutes? I am wtriting typings for the filter plugin and have one weired error popping up in the d.ts file

PR #961, opened 2019-02-16
Added missing 'clone(parent)' overload to .d.ts file

Perlkonig:
Applied against 2.x branch.


Fuzzyma:
> Applied against 2.x branch.

That would have been my first question :D

PR #963, opened 2019-02-17
Fix typo in type file for Polyline and Marker

Perlkonig:
Applied against 2.x

### DIFF
--- a/svg.js.d.ts
+++ b/svg.js.d.ts
@@ -460,7 +460,7 @@ declare namespace svgjs {
         marker(position: string, width?: number, height?: number, block?: (marker: Marker) => void): Marker;
         marker(position: string, marker: Marker): Marker;
     }
-    interface PolyLine {
+    interface Polyline {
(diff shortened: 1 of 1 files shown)


Fuzzyma:
Are you sure about that one? I thought Polyline is Polyline and not PolyLine.

Perlkonig:
I'm no expert, as you know!! 😄 

Here's the code that wasn't working:

```typescript
const line = notes.polyline(points.join(" ")).stroke({width: 1.5, color: "black"});
line.marker("end", markerArrow);
line.marker("start", markerCircle);
```

The error was `Property 'marker' does not exist on type 'PolyLine'. Did you mean 'masker'?`.

So I opened the `.d.ts` file and saw that everything was `PolyLine` except for the one that defined `marker()`, so I thought it was a typo and corrected it. Everything compiled fine after that.

Fuzzyma:
Hm okay. To stay consistend, could you rename all occurences to Polyline?

Fuzzyma:
thanks!

Issue #964, opened 2019-02-21
timeline.finish() is not working 


michAtEl:
# Bug report

## Fiddle
Try this codepen:
https://codepen.io/michAtEl/pen/KJYBzP

Uncomment the last line (//timeline.finish()) and run it again. The blue box reaches the wrong position (if my understanding of the API of timeline is correct).

## Explanation
Timeline.finish() calls Timeline.getEndTime(). getEndTime() only looks at the last Runner in timeline._runners  to calculate the endtime. In my example, the last element in this array does not have the latest endtime; in my example, the first element in this array has the latest endtime.

## How to fix
I think it could be enough to loop over all runners in getEndTime() and find the runner with the latest endtime.
As long as it is not fixed: I make sure that the runner with the latest endtime is in the last position.

Fuzzyma:
Good catch. It also seems that `getEndTime` is not time critical so looping over all runners could be a fix.
However, getEndTime is also used in schedule and for that it has to work like it is atm. So a replica of the old method is needed

PR #966, opened 2019-02-23
bugfix for issue #964

michAtEl:
In Timeline.getEndTimeOfTimeline() I check weather runnerInfo is truthy: 
      var duration = runnerInfo ? runnerInfo.runner.duration() : 0
      var startTime = runnerInfo ? runnerInfo.start : 0

I'm not sure weather this is required; this might be enough:
      var duration = runnerInfo.runner.duration()
      var startTime =runnerInfo.start

I do these checks because they are done in Timeline.getEndTime()

### DIFF
--- a/spec/SpecRunner.html
+++ b/spec/SpecRunner.html
@@ -105,5 +105,6 @@
   <script src="spec/animator.js"></script>
   <script src="spec/runner.js"></script>
   <script src="spec/queue.js"></script>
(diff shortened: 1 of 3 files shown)


michAtEl:
Hi,
is something wrong with this pull requets?
Michael

Fuzzyma:
nope^^

PR #969, opened 2019-02-26
Add typescript test and update typescript definition

cmidt-veasna:
Update typescript definition to include SVG Attribute and add typescript test. The typescript test has two phrase, first the compilation of typescript into javascript and second **webpack** will bundle all javascript and perform a test. To perform this step run the command `npm test:types` or `yarn test:types`.


svg.js.d.ts
```diff
@@ -1078,6 +1139,7 @@ declare module "@svgdotjs/svg.js" {
     class Svg extends Container {
         constructor(svgElement?: SVGSVGElement);
         constructor(id: string);
+        constructor(domElement: HTMLElement);
```
Fuzzyma: This one is not true. The `<svg>` constructor can only take a node or an attributes array. Dont mix this up with the `SVG()` method which can take all these.

cmidt-veasna: @Fuzzyma Yep, you're right the constructor should different from SVG().

cmidt-veasna:
@Fuzzyma I agree, replicated ALL test cases does not seem to be a good practice however I can't figure out the better way than this way. The purpose of this test is to make sure that typescript definition file is well defined and align with Javascript implementation (Maybe not require all to be aligned). The typescript is just a compiler and output a javascript so it seem impossible to incorporate with existing javascript test. On the other hand, the typescript test required 2 step, first, we need to ensure compiler does not raise any error against `.d.ts` file and second the output of the javascript should work as expected (Same way as javascript test). So this is reason why I replicate all the test and I can't figure out how to incorporate the output result with existing javascript test. **_ To be fair, I don't think a library that contain only typescript definition (`.d.ts` file) is required to have typescript test_**, it's nice to have but we have to endure this issue.

I attempted to update the definition file since 3.x is out by looking into Change Log, Document and finally when deep into Javascript source code as I don't feel typescript definition is aligned with Javascript API. With all of these step, I still not sure if the output result of typescript is work the same as we wrote in javascript.

@Fuzzyma as for question:
> Would it be possible, to have esm imports and get rid of all old test cases without basically screwing up the tests while making typescript tests possible?
I think that probably possible, however the typescript test and definition file might need to be updated.
> if it is possible to use rollup for typescript bundling
I'm not really sure but it should be possible. I just need to take sometime to try it.

@Fuzzyma with latest update of the definition file, it seem that the definition is close to javascript API. I think the test is not really needed right now and we should wait until you finish the test in es6 and perhaps we might find better solution to incorporate the typescript with existing test.

cmidt-veasna:
@Fuzzyma I will close this pull request for now. Will wait for a update to the existing javascript test.

Fuzzyma:
Sorry for not answering. I didn't do it directly and forgot about it (as always). I wanted to ask you if you could make a PR for the fixes in the d.ts file only.
We could then move your typescript tests into another branch where it can sit for a while until we decide what to do with it (so it doesn't get lost).

Would that be OK?

cmidt-veasna:
@Fuzzyma sure, I will push the update tonight.

Fuzzyma:
@cmidt-veasna I came back to this PR because the first time now I am using typescript for a project and I wanted to find out which amazing person created this super huge d.ts file for svg.js.
**This is amazing! Just wanted to thank you again for that.**
The only problem I have is, that I dont even understand half of the magic going on there :D. But I find more and more small mistakes here and there which I am hopefully able to fix myself :).
I also found a few more broader ones. Actually I am finding them right now because I am looking at the code while writing this :D. E.g. the `parent()` method also accepts query selectors and `putIn` accepts not only container elements but others too (because all shapes can have a title, desc and metadata element as child and animation elements, too even though we dont support it).

After using typescript I feel like having tests in typescript could actually make the libraray more stable in general. However, our tests are not up to date at this point anyway (it is SO much work and my motivation for that is kinda limited).

The text went longer than I expected. Actually I only wanted to thank you but I went full mode again - sorry :D

PR #970, opened 2019-02-28
Updated ease type in svg.js.d.ts

manospasj:
Updated animate method to accept a custom ease function as per the docs (https://svgjs.com/docs/2.7/animating/#easing)

### DIFF
--- a/svg.js.d.ts
+++ b/svg.js.d.ts
@@ -1127,8 +1127,8 @@ declare module "@svgdotjs/svg.js" {
 
         addClass(name: string): this;
         after(element: Element): Element;
-        animate(duration?: number, ease?: string, delay?: number): Runner;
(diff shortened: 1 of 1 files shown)


Fuzzyma:
Thanks for your contribution. It seems like the old and the new code is wrong for v3 :D.
The animate method can take the following parameters:

```js
el.animate(duration, delay, when)
// or
el.animate(obj)
// or
el.animate(controller, delay, when)
```
Duration and delay are numbers. `when` is an enum with the values `now, after, absolute, relative`. It defaults to `after` (which is `now` when there is no animation before this one).

The obj can hold all these values as object keys.
The controller is that, a contoller which controls the animation. There is a class for that.

manospasj:
Oh, I thought that this branch was for version 2. Sorry about that, I will create a new pull request in the correct branch.

PR #1010, opened 2019-06-13
transform() works with Matrix, not with MatrixExtract or MatrixTransformParam

showdep:


Fuzzyma:
Hm, not really true. transform can handle transform params as well as matrices. It returns TransformData + the a-f of the matrix

Fuzzyma:
Sorry for not coming back on this. The first issue is fixed but transform still returns transform data and teh matrix. not only a matrix

PR #1065, opened 2019-12-12
Fixes Element.animate typing

### LINKED ISSUE #1064: Element.animate typing parameters incorrect in 3.0
# Bug report

The TypeScript typing for the Element.animate method should match the Runner.animate method.

## Fiddle

N/A

## Explanation

- Using TypeScript, the Element.animate method should allow the correct parameters `duration, delay, when` in 3.0.
- Using TypeScript, the Element.animate method currently expects `duration, ease, delay`. Attempting to use the correct parameters results in a TypeScript compiler error.

mradcliffe:
Resolves #1064

### DIFF
--- a/svg.js.d.ts
+++ b/svg.js.d.ts
@@ -1169,8 +1169,7 @@ declare module "@svgdotjs/svg.js" {
 
         addClass(name: string): this;
         after(element: Element): Element;
-        animate(duration?: number, ease?: string, delay?: number): Runner;
-        animate(info: { ease?: string; duration?: number; delay?: number }): Runner;
+        animate(duration?: number, delay?: number, when?: string): Runner;
         delay(by: number, when?: string): Runner
         attr(): any;
         attr(name: string, value: any, namespace?: string): this;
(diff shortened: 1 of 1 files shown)


svg.js.d.ts
```diff
@@ -1169,8 +1169,7 @@ declare module "@svgdotjs/svg.js" {
 
         addClass(name: string): this;
         after(element: Element): Element;
-        animate(duration?: number, ease?: string, delay?: number): Runner;
-        animate(info: { ease?: string; duration?: number; delay?: number }): Runner;
+        animate(duration?: number, delay?: number, when?: string): Runner;
```
Fuzzyma: Albeit this is correct, it also allowed for `animate({object})` like:

```
animate({
  duration: ...,
  when: ...,
  delay: ...,
  swing: ...,
  times: ...,
  wait: ...,
})
```

`swing, times and when` are known from the `loop` method

mradcliffe: Thank you for the review! :+1: I was having trouble finding this signature in the code, but I took another look and I was able to find it in sanitise :-)

I changed both the function signatures to use the `TimeLike` type that was already defined.

PR #1092, opened 2020-03-11
Add typing for persist method in Timeline and Runner

josepaiva94:
Added types for persist method in Timeline and Runner


Fuzzyma:
this one is already fixed locally on my laptop. Just wait a little longer :)

josepaiva94:
Ok :) will close then!

PR #1106, opened 2020-05-07
…

pragdave:
It was difficult to use the zero-parameter form of `Timeline.schedule()` with TypeScript. The `svg.js.d.ts` file didn't include the return type for a list of runner infos, and adding it still made it harder than it should be to iterate over it type safely. 

In this commit I:

1. Added the type information for `ScheduledRunnerInfo`, and updated the Timeline types to include it as a return type for `schedule`

2. Added a new function to Timeline, `getSchedule`, that simply returns the runner list. This seems to be cleaner than having the original `schedule` that can return two wildly different things.

I didn't remove the old functionality, so `schedule()` is still compatible. I suggest we might want to deprecate that old form at some point.

### DIFF
--- a/src/animation/Timeline.js
+++ b/src/animation/Timeline.js
@@ -44,17 +44,10 @@ export default class Timeline extends EventTarget {
     this._stepImmediate = this._stepFn.bind(this, true)
   }
 
-  // returns information on each runner in the timeline.
-  // (start, duration, end, and runner)
-
-  getSchedule () {
-    return this._runners.map(makeSchedule)
-  }
-
   // schedules a runner on the timeline
   schedule (runner, delay, when) {
     if (runner == null) {
-      return this.getSchedule()
+      return this._runners.map(makeSchedule)
     }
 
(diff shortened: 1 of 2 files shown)


Fuzzyma:
Hm, it is quite common in svg.js to get and set a value with the same function. However, you dont write `this | Number` as return type but instead you have an overload. Otherwise you get back a union type and typescript doesnt know if its the one or other type. So writing it seperately would actually solve your issue:

```ts
        schedule(runner: Runner, delay?: number, when?: string): this
        schedule(): ScheduledRunnerInfo[]
```

I also made the runner non-optional (which is correct). It is hard to keep the ts file always up to date but I try my best :).

As for the PR: I would rather not introduce a new method here. I would accept the PR without the new method and with the suggested changes I made in this post :)

pragdave:
re the union: Oh, of course

pushed

Fuzzyma:
Perfect! Thank you!

PR #1111, opened 2020-05-20
Add optional string type to Runner.ease

acedrew:
The ease method on the runner takes the same constructor strings as the Ease class, so you can do things like element.animate(1000).ease('-').rotate(360).loop() 
Currently typescript throws an error saying that ease only accepts a Function as an argument.


Fuzzyma:
**Congratulations** - you opened the 1111th issue on this repo :tada:  :partying_face:

And well yes - you are correct. Thank you! :)

PR #1116, opened 2020-05-27
Better typing for css getter/setter

Glandos:
CSSStyleDeclaration contains readonly rule, so I excluded them, otherwise it doesn't make sense
I didn't touch the Runner `css()` because I don't know the exact behavior of it.


Fuzzyma:
nice!

PR #1120, opened 2020-05-28
Allow newline to have a count

pragdave:
Here's that change I mentioned. I updated the tests, and they all pass.  I also tried it on my code based, and it looks good.

There's one failing test, but it also fails in master:

~~~
Firefox 76.0 (Mac OS 10.14) textable.js x() sets the value of all lines FAILED
        Expected 201 to be 200.
        <Jasmine>
        @spec/spec/modules/core/textable.js:27:26
        each@src/elements/Dom.js:84:13 <- src/elements/Dom.js:11:200
        @spec/spec/modules/core/textable.js:26:12
        <Jasmine>
        Expected 201 to be 200.
        <Jasmine>
        @spec/spec/modules/core/textable.js:27:26
        each@src/elements/Dom.js:84:13 <- src/elements/Dom.js:11:200
        @spec/spec/modules/core/textable.js:26:12
        <Jasmine>
~~~

I'm doing the documentation now.


Fuzzyma:
I think i will review this when a new major version comes out and close it for now

PR #1123, opened 2020-05-30
add foreignObject to types, and get List working

pragdave:
I found that I was getting type errors with Lists when it was an interface, but if I changed it to a class and tightened up the callback function and return types it was OK.

I also added the ForeignObject class and helper function in Containers


svg.js.d.ts
```diff
@@ -1382,6 +1385,13 @@ declare module "@svgdotjs/svg.js" {
         target(): string;
     }
 
+
+    // ForeignObject.js
+    class ForeignObject extends Element {
+        constructor(node: HTMLElement, attrs?: any)
+        add(element: Dom) : ForeignObject
```
Fuzzyma, line 1392: Isnt `add` already covered in the Dom interface?

svg.js.d.ts
```diff
@@ -548,13 +548,15 @@ declare module "@svgdotjs/svg.js" {
...
+
     // List.js
-    interface List<T> extends BuiltInArray<T> {
+    class List<T> extends BuiltInArray<T> {
         // I have no clue how to deal with this
         // [key: string]: (...arg0: any[]) => List<T>
         // [key: string]: () => List<any>
-        each(fn: Function): void
-        each(...args: any[]): void
+        each(fn: ListEachCallback<T>): T[]
```
Fuzzyma, line 558: Actually this  method can take a function or a string and it doesnt return always an array of T. Whatever the function return you pass to it (as function or string name) will be maped into the returned array.
This is pretty much impossible to type -.-. But it always returns a List!

```suggestion
        each(fn: ListEachCallback<T>): List<any>
```

e.g. `list.each('fill')` will return a new list with contains all the fill colors of the elements in the list because it acts as getter. Or `list.each('animate')` will contain lots of runners.

However, the string version actually needs to go onto the next line...

svg.js.d.ts
```diff
@@ -548,13 +548,15 @@ declare module "@svgdotjs/svg.js" {
...
+        each(...args: any[]): T[]
```
Fuzzyma, line 559: ... here :D 

```suggestion
        each(fnName: string, ...args: any[]): List<any>
```

svg.js.d.ts
```diff
@@ -548,13 +548,15 @@ declare module "@svgdotjs/svg.js" {
         to(a: any): Morphable;
     }
 
+    type ListEachCallback<T> = (el: T, index: number, list: T[]) => T
+
     // List.js
-    interface List<T> extends BuiltInArray<T> {
+    class List<T> extends BuiltInArray<T> {
         // I have no clue how to deal with this
         // [key: string]: (...arg0: any[]) => List<T>
```
Fuzzyma, line 556: In case you have an idea how to make this part work... its basically all methods of everything on the List e.g. `list.move(10, 10).animate().move(40, 40)`

svg.js.d.ts
pragdave: The one in Dom is typed as

        add(element: Element, i?: number): this;

but ForeignObject accepts a Dom...

As it stands, it isn't really typeable, because the two `add` functions do different things. I can't think of a way around it apart from changing the method name in ForeignObject to something like `addHTML`.

I have to admit I don't fully understand how the source works here:

~~~ js
export default class ForeignObject extends Element {
  constructor (node, attrs = node) {
    super(nodeOrNew('foreignObject', node), attrs)
  }
}

registerMethods({
  Container: {
    foreignObject: wrapWithAttrCheck(function (width, height) {
      return this.put(new ForeignObject()).size(width, height)
    })
  }
})
~~~

The constructor takes a node, but it's called with nothing.

svg.js.d.ts
pragdave: I don't think returns a List: isn't it just a lower-case array?

How about this:

~~~ js
    class List<T> extends BuiltInArray<T> implements ElementAlias {
        each(fn: ListEachCallback<T>): any[]
        each(name: string, ...args: any[]): any[]
        toArray(): T[]
    }
~~~

Does implementing `ElementAlias` cover the interface? Would we need Runner as well?

svg.js.d.ts
pragdave: See above.

svg.js.d.ts
Fuzzyma: Ah ok, yes, in that case foreignElement should get its own `add` typings. But dont forget the second parameter then :).

For the constructor: You CAN create an ForeignObject-Object (lol) by passing in a node. It will then use that node instead of creating a new one. Thats why there is the `nodeOrNew` call which gets passed the node and the element to create if node is not defined.

The attrs can be passed to any constructor to allow to directly set attributes on the element e.g. `new Rect({width: 10, height: 20})`

svg.js.d.ts
Fuzzyma: No, it does indeed return a new list. The code is `this.map(fn)` which will create a new List. You can try it for yourself.

ElementAlias is not an interface. It is just all svg.js types as a union. You would need an interface with ALL the methods which are available on any element of svg.js.

svg.js.d.ts
```diff
@@ -548,13 +548,12 @@ declare module "@svgdotjs/svg.js" {
         to(a: any): Morphable;
     }
 
+    type ListEachCallback<T> = (el: T, index: number, list: T[]) => any
```
Fuzzyma, line 551: Just saw this: The third paramater `list` will be a List, too. So I think it has to be `list: List<T>`

pragdave:
> ElementAlias is not an interface. It is just all svg.js types as a union.
> You would need an interface with ALL the methods which are available on any
> element of svg.js.
>

But if we said:

~~~
class ForeignObject extends ... implements Rect {
}
~~~

then wouldn't that means that it would be typed to include all the methods
in Rect? So wouldn't the union type say 'all the methods in any of
these..."?

svg.js.d.ts
```diff
@@ -1382,6 +1388,13 @@ declare module "@svgdotjs/svg.js" {
         target(): string;
     }
 
+
+    // ForeignObject.js
+    class ForeignObject extends Element {
+        constructor(node: HTMLElement, attrs?: any)
```
Fuzzyma, line 1394:
```suggestion
        constructor(node?: SVGForeignObjectElement, attrs?: object)
        constructor(attrs?: object)
```

This is wrong in other places, too. But thats how it should be

svg.js.d.ts
```diff
@@ -1382,6 +1388,13 @@ declare module "@svgdotjs/svg.js" {
         target(): string;
     }
 
+
+    // ForeignObject.js
+    class ForeignObject extends Element {
+        constructor(node: HTMLElement, attrs?: any)
+        add(element: Dom, attrs?: any) : ForeignObject
```
Fuzzyma, line 1395: The `element` can actually be anything from a html string over a html node to a svg.js object.
Hard to type...
And attrs is completely wrong here. `add()` takes an optional position `i` as second parameter

svg.js.d.ts
```diff
@@ -80,14 +80,14 @@ declare module "@svgdotjs/svg.js" {
...
 
-    let easing: {
-        '-'(pos: number): number;
-        '<>'(pos: number): number;
-        '>'(pos: number): number;
-        '<'(pos: number): number;
-        bezier(x1: number, y1: number, x2: number, y2: number): (t: number) => number;
-        steps(steps: number, stepPosition?: "jump-start"|"jump-end"|"jump-none"|"jump-both"|"start"|"end"): (t: number, beforeFlag?: boolean) => number;
-    }
+    // let easing: {
```
Fuzzyma, line 83: This easing object is also exported so it should actually be an export

Fuzzyma:
I honestly have no idea if you can implement Unions. Also: if you implement them you must actually write them out in the typings and thats what I wanted to avoid.

**Note to myself**: I need to add SVGForeignObjectElement to the typeMappings in the d.ts file

svg.js.d.ts
pragdave: Added back...

pragdave: So let me ask a consultant-type question: what is the use case for ForeignObject, not in theory but in practice. What hole does it help fill?

Right now is will accept basically anything and its dog; maybe it doesn't need to. Perhaps _all_ you need to support is HTMLElement. What would be lost if that were the case?

pragdave:
So I just spent a couple of hours messing around with trying to get a safe typing for List. I've learned a lot about keyof, indexing, and mixins, but I don't seem to be able to create something as dynamic as we need. It's on my back burner of things to think about while asleep.

In the meantime, this PR has turned into something of a mess. I'd like to get at least some of it merged, just because it fixes a lot of errors I'm seeing here. This might be a case where pragmatism beats perfection :)

Would it be best to close this PR and open another? Or do you know what you're happy to merge?

svg.js.d.ts
Fuzzyma: ForeignObject can be filled with basically any Node you want. It is for content which is NOT svg. So it can be html, mathml, xml - whatever you want.
Since svg.js also works in standalone svgs and on the server, it basically needs to support anything which can be intepretet by any interpreter which exist or might exist.

But were is this question coming from? Because of the `add()` method?

Fuzzyma:
> So I just spent a couple of hours messing around with trying to get a safe typing for List. I've learned a lot about keyof, indexing, and mixins, but I don't seem to be able to create something as dynamic as we need

Yes, I honestly thought, that a string-index-key as I tried would allow this but you can either have that OR other methods. Because all the array methods have to follow the string-index signature, too. Its stupid. Thats why it cannot really be typed and that is really unfortunate.

> In the meantime, this PR has turned into something of a mess. I'd like to get at least some of it merged, just because it fixes a lot of errors I'm seeing here. This might be a case where pragmatism beats perfection :)

I honestly lost track a bit :D. I will have a look which conversation can be resolved and will do a last review of what needs to be changed explicitely. Maybe I just change it myself if that works over the browser. And then we can merge :)

svg.js.d.ts
Fuzzyma:
```suggestion
        add(element: Dom, i?: number): ForeignObject
```

I see there is a lot of mess going on. Have to dig deeper

PR #1126, opened 2020-06-02
Added exports from window.js

pjbatista:
Added `getWindow`, `restoreWindow`, `saveWindow`, and `withWindow` to the exports of "main.js" and to the type declarations in "svg.js.d.ts".

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1222,9 +1222,9 @@
       }
     },
     "acorn": {
-      "version": "7.1.1",
-      "resolved": "https://registry.npmjs.org/acorn/-/acorn-7.1.1.tgz",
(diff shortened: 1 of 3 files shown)


Fuzzyma:
thanks!

bzuker:
Hi! Is there any chance to do a patch release for these changes? I need to use `registerWindow` but it's currently not exported in version `3.0.16`. Thanks!

bzuker:
> @bzuker registerWindow IS exported. Only the other window functions are not

Maybe I'm looking in the wrong place, but I'm checking here: https://github.com/svgdotjs/svg.js/blob/3.0.16/svg.js.d.ts
And I don't see it exported there.

I notice the function is exported here https://github.com/svgdotjs/svg.js/blob/3.0.16/src/main.js (which is probably what you mean)

However, I could work around it doing: 

```
declare module "@svgdotjs/svg.js" {
  const registerWindow: Function;
}
```

Thank for your quick response!

PR #1128, opened 2020-06-14
Fix types for List and ForeignObject

josepaiva94:
Fixes typescript errors List and ForeignObject introduced in #1123


Fuzzyma:
hey! thanks for the PR. Unfortunately it seems like you changed every single line of the whole file (line endings maybe?). I cannot accept the PR like that. Can you fix this?

PR #1175, opened 2020-12-12
fix arc flags parsing

mlwilkerson:
This PR proposes a fix for parsing Elliptical Arc commands in SVG paths. The parser has been failing to parse some valid arc commands. For example, the [grammar](https://www.w3.org/TR/SVG/paths.html#PathDataBNF) allows for whitespace/comma to be omitted following one of the arc flags.

### DIFF
--- a/spec/spec/utils/pathParser.js
+++ b/spec/spec/utils/pathParser.js
@@ -102,6 +102,32 @@ describe('pathParser.js', () => {
         [ 'L', 0.3, 0.3 ],
         [ 'Z' ]
       ])
+
+      // "a" commands without optional whitespace around the flag params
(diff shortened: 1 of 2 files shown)


Fuzzyma:
Hm, I think I had that and couldnt find it in the grammar. Can you add another testcase for 3 consecutive zeros (or similar) for flag, flag and following number?
Thanks for collaborating!

mlwilkerson:
Ok, added some more test cases. All passing. (Though there was already the one failing test case at `spec/modules/core/textable.js:27`  in `master` before I branched.)

Fuzzyma:
good to know :D. And thank you!!

PR #1179, opened 2020-12-17
typos

ordago:
All the changes are in either comments or test messages.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -6,7 +6,7 @@ When contributing to this repository, please first discuss the change you wish t
 
 ## Pull Request Process
 
-When you want to make contriubtions to the project, the process is pretty simple:
+When you want to make contributions to the project, the process is pretty simple:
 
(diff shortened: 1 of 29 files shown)


Fuzzyma:
Oh wow! Did you use software or did you really do this one by one? :D

ordago:
Hi @Fuzzyma 
I used Phpstorm, and then went over the results manually since the names of variables, constants... give a lot of false positives.


I also saw these but didn't touch it because it was in code:

https://github.com/svgdotjs/svg.js/blob/4f4dc2afa62cb899a5cae72c359150f8adf27edb/spec/spec/modules/optional/sugar.js#L28

https://github.com/svgdotjs/svg.js/blob/4f4dc2afa62cb899a5cae72c359150f8adf27edb/spec/spec/modules/optional/sugar.js#L82

These should be `fixtures` instead of `fictures`, right?

Fuzzyma:
@ordago wow, thanks! Yes, it should be fixtures. For the test, it doesn't really matter but ofc it would be better to be correct rather then wrong :D

PR #1219, opened 2021-09-22
node or new

visionm45:
edit nodeornew in order to behave consistently across different contexts.

note. npm automatically updated the package lock file and I pushed it without realizing so it would probably not be desired to merge that.


Fuzzyma:
just reset the package lock to the old state and commit it again :)

src/utils/adopter.js
```diff
@@ -42,7 +42,7 @@ export function makeInstance (element, isHTML = false) {
 }
 
 export function nodeOrNew (name, node) {
-  return node instanceof globals.window.Node ? node : create(name)
+  return (node && 'ownerDocument' in node && node instanceof node.ownerDocument.defaultView.Node) ? node : create(name)
```
Fuzzyma, line 45: probably checking `node.ownerDocument` is better here

visionm45: oh right. you got me thinking what is the difference anyway it seems to be that 

> [If an object has a property whose value is undefined, then comparing the property with undefined will return an incorrect result.](https://www.javascripttutorial.net/object/javascript-check-if-property-exists/)

But when checking node.ownerDocument there seems to be no problems and the tests pass just the same so I agree.

visionm45:
I'm a bit new to using git so sorry if some of the things that I did seem a bit strange.  Anyway I think that now the only difference in this branch is in the function nodeOrNew

src/utils/adopter.js
Fuzzyma: Well in javascript `undefined` usually means "not defined" on the object. Even if you set the property to undefined, that's what you often want to communicate. Otherwise, you use `null`. The `in` check is an escape hatch in case you REALLY need it. Usually checking just for a truthy value is enough.
I also realized that we don't really need the `instanceof` check anymore because the whole purpose of this check is to distinguish a Node from a plain object.
But I guess its fine like that. Thanks!

Fuzzyma:
No problem. I can just squash everything into one commit so nobody sees all the commits

src/utils/adopter.js
visionm45: Okay. I was thinking the same thing about using `instanceof` actually.  The main reason I left it was because it was what was originally being done.

Fuzzyma: We'll leave it for now :)

PR #1231, opened 2021-11-05
findOne can return null

Glandos:
This can be a breaking change. But it will also fix the contract of the method.


Fuzzyma:
People who used that function had to catch that case anyway. So I don't consider this a breaking change (also I don't consider any typescript update a breaking change anyway). Thanks for sporting!

Issue #1232, opened 2021-11-10
Typings for dmove missing in Runner class


adamyakes:
# Bug report

When using TypeScript, the Runner class is missing the method `dmove(dx, dy)`.

## Fiddle

N/A

## Explanation

Getting TS error "Property 'dmove' does not exist on type 'Runner'".

PR #1233, opened 2021-11-10
Adds dmove to type definition for class Runner

adamyakes:
Current typings for `Runner` do not include the `dmove` method defined [here](https://github.com/svgdotjs/svg.js/blob/master/src/animation/Runner.js#L885). This PR adds that definition.

Created #1232 for this PR.


Fuzzyma:
Simple and efficiant :D - thanks!!

PR #1234, opened 2021-11-16
Fix throwing when call svg.ungroup()

OpenGG:
Current status: `svg.ungroup()` throws with confusing message "Uncaught TypeError: parent.screenCTM is not a function".
Fix: Early throws with determine message.

### DIFF
--- a/src/elements/Container.js
+++ b/src/elements/Container.js
@@ -13,6 +13,9 @@ export default class Container extends Element {
   }
 
   ungroup (parent = this.parent(), index = parent.index(this)) {
+    if (this.isRoot()) {
+      throw new Error('Can not ungroup svg root element')
+    }
     // when parent != this, we want append all elements to the end
     index = index === -1 ? parent.children().length : index
 


Fuzzyma:
Another possibility would be to just call flatten instead. But that might be confusing.
We really rarely throw errors in the lib because validating all user input would quickly bloat the code. How did you discover this?

OpenGG:
@Fuzzyma

OpenGG:
@Fuzzyma 
Am really just newbee digging around.
If `ungroup()` not supposed to be called on svg root, maybe it should be moved to Group class?

Fuzzyma:
Well you can call it on a nested svg. Also there are other container elements that you can ungroup. Basically everything that can have children (defs for example)

OpenGG:
@Fuzzyma 
Then what should be done with this issue in your opinion?
Option 1: `.ungroup()` can be call with "basically everything"
In this case, early throwing or polymorphism? Both requires special handling with different element.
Option 2: Limit `.ungroup()` to <g>
Relatively simple, may break `flatten()`.
Option 3: Do nothing
Calling `.ungroup()` with <svg> cause throwing, other elements may too.

Fuzzyma:
Ungroup can ONLY be called for container elements. The only option is leave it or implement your fix. However your fix will break for other elements than svg because they don't have the isRoot method. So I guess we leave it for now

PR #1236, opened 2021-11-23
Fix #1235

### LINKED ISSUE #1235: parents(selector) does not find ancestors by selector
# Bug report

`parents(selector)` does not behave as expected. It does not always terminate at the closest matching ancestor; instead, this behaviour only occurs if the *first element matching the selector in the document* is a direct ancestor of the calling element.

## Fiddle

https://jsfiddle.net/gormster/2xw7sape/

### Expected behaviour
`parents(selector)` should return the ancestors of the current element up until an element that matches `selector`, as documented. In the supplied fiddle, both circles should turn green.

### Actual behaviour
The returned list of elements only terminates if the first element in the document that matches `selector` is a direct ancestor of the calling element. In other words, if `element.root().findOne(selector)` is in `element.parents()`, then `element.parents(selector)` works (almost) as expected. However:

* If there are multiple elements in the parent chain that match the selector, `parents(selector)` matches the *furthest* element, not the closest one (https://jsfiddle.net/gormster/erhwjv1c/)
* If there are multiple elements in the document that match the selector, and the first such element is not in the parent chain, `parents(selector)` returns all parents including the elements in the containing HTML document (https://jsfiddle.net/gormster/2xw7sape/)
* If there are no elements in the document that match the selector, `parents(selector)` throws a TypeError (https://jsfiddle.net/gormster/c7k23vt0/)

gormster:
Fixes the bugs in `Element.parents()` function.

This also introduces a minor change to functionality, but it's a change that previously threw a TypeError, so I don't think it can be considered a "breaking change". It's also more-or-less what the docs say this function does, although they're not specific.

**This function no longer searches outside the containing SVG document; if it reaches `this.root()` without finding a matching element, it returns `null`.**

This is, I think, what most devs would expect. If I call `parents('.foo')`, I expect the last element in that array to have the class `foo`. I'm confident most devs would agree.

### DIFF
--- a/spec/spec/elements/Element.js
+++ b/spec/spec/elements/Element.js
@@ -188,6 +188,20 @@ describe('Element.js', function () {
       expect(rect.parents('.foo')).toEqual([ group3 ])
       expect(rect.parents('.test:not(.foo)')).toEqual([ group3, group2, group1 ])
     })
+
+    it('returns null if the passed element is not an ancestor', () => {
+      const canvas = SVG().addTo(container)
+      const groupA = canvas.group().addClass('test')
+      const group1 = canvas.group()
+      const group2 = group1.group()
+      const group3 = group2.group()
+      const rect = group3.rect(100, 100)
+
+
+      expect(rect.parents('.does-not-exist')).toEqual(null)
+      expect(rect.parents('.test')).toEqual(null)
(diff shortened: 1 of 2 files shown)


Fuzzyma:
Good catch. This method is used so rarely that bugs live forever before they are found (and debugged AND fixed). Thank you very much!

PR #1251, opened 2022-01-27
Performance: don't create a double clone in point

Glandos:
The Point instance is already created, so don't create a new one.

Edit: also don't clone the newly created Matrix from `getScreenCTM`


Fuzzyma:
Good catch!

PR #1258, opened 2022-02-20
fix: Don't same variable name as variable function

lpatiny:
When using this library in Safari (OsX or iPad) we may encounter the error:

'duplicate parameter in function with default parameter values'

This is related to the use of a variable name that is the same as the function and similar fix were applied in other libraries like:
https://github.com/jupyterlab/jupyterlab/pull/7993/files

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,4 @@
+{
+  "semi": false,
+  "singleQuote": true
+}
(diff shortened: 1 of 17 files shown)


Fuzzyma:
Is this a bug in Safari?

Issue #1277, opened 2022-08-09
maskWith quotation approach


HarryFaulkner:
Not sure if this is a bug or something to do with the SVG versioning... but for anyone experiencing issues importing masked SVG's into the applications I mention below, this is likely _your_ issue.

When using the .maskWith() command, the mask is double quoted:

`<g> mask="url("#SvgjsMask1002")" </g>`

This approach works fine in the browser and _most places_. But when downloading the SVG and importing to something like Adobe _illustrator_ or _Inkscape_ it fails to render the mask.

The following line, wraps the mask in single quotations, and a supplemental double quotation for the internal #id.

https://github.com/svgdotjs/svg.js/blob/4524c20b48abe08e69c25a7382a289592d147496/src/elements/Mask.js#L47

I've adjusted my local version to the following (which fixes the issue for me):
`return this.attr('mask', 'url(#' + masker.id() + ')');`

I don't know a great deal about the SVG versioning (changes to syntax that may have happened over time), this may be correct for a given version. The current documentation seems to suggest double quotes are not required:
https://developer.mozilla.org/en-US/docs/Web/SVG/Element/mask

If this is intended behaviour then of course feel free to close this.

Fuzzyma:
To me this sounds like a bug in the parser of Adobe or Inkscape - or its not allowed from the standard and the browser still works.

Nevertheless, yes, we should use single quotes if they get around that issue

HarryFaulkner:
> To me this sounds like a bug in the parser of Adobe or Inkscape - or its not allowed from the standard and the browser still works.
> 
> Nevertheless, yes, we should use single quotes if they get around that issue

Agreed, dam illustrator, they should be able to _at least_ handle what a measly browser can 😄 

PR: https://github.com/svgdotjs/svg.js/pull/1278

PR #1278, opened 2022-08-09
1277: remove double quotes from maskWith url() 

HarryFaulkner:
https://github.com/svgdotjs/svg.js/issues/1277


Fuzzyma:
We have multiple spots with the same Syntax in svg.js. If we change it, we should change it anywhere. e. G. Clip-path.

Also: it seems like you removed the quote all together. Wouldn't it be better to replace it with a single quote instead?

HarryFaulkner:
> We have multiple spots with the same Syntax in svg.js. If we change it, we should change it anywhere. e. G. Clip-path.
> 
> Also: it seems like you removed the quote all together. Wouldn't it be better to replace it with a single quote instead?

**Change it everywhere**
Sure, will add it in the other areas also. 

**Replace quote with single quote**
This is the syntax used by the official documentation, note that there are no inner quotes
`  <circle cx="50" cy="50" r="50" mask="url(#myMask)" />`
[https://developer.mozilla.org/en-US/docs/Web/SVG/Element/mask](url)

HarryFaulkner:
Any chance of this getting merged?

rossburton:
This is breaking my SVGs that I want to load into Affinity, so yeah merge please :)

Fuzzyma:
sorry for the long wait. my bad!

Issue #1307, opened 2023-08-13
No Fragment declaration in svg.js.d.ts


we125182:
![image](https://github.com/svgdotjs/svg.js/assets/17947588/3db455eb-611f-4554-bc8d-1150889665ee)

`svg.js.d.ts` don't export Fragment

**expect**
add `Fragment` class declaration

Fuzzyma:
Are you open to create a PR?

we125182:
OK，I’d like try that.

PR #1309, opened 2023-08-14
fix: add Fragment declaration in svg.js.d.ts

we125182:
this pr is to fix issue #1307 .

### DIFF
--- a/svg.js.d.ts
+++ b/svg.js.d.ts
@@ -1204,8 +1204,7 @@ declare module '@svgdotjs/svg.js' {
     height: number
   }
 
-  // Fragment is special Container: maybe there is a way to avoid this duplication?
-  class Fragment extends Dom {
+  class Containable {
     circle(size?: NumberAlias): Circle
     circle(size: number, unit: number): Circle
     clip(): ClipPath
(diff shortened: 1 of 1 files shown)


Fuzzyma:
Would it make sense to let fragment simply extend from Container to avoid duplication?

we125182:
Actually, I thought in this way. After read source code, find that Container extends Element, but Fragment extends Dom. if Fragment extends Container, it'll has Element methods, but Fragment doesn't have methods of Element

Fuzzyma:
mh, maybe you can add an interface "Containerable" which has all the methods and both fragment and container extend from it

we125182:
I tried the Interface way. But implements `Containerable` Interface, you need define same medthods too. in this way, need define same method three times(`Containerable`, `Fragment`, `Container`). Certanlly use interface can ensure `Container` and 'Frament` has same Constuct Element methods.

image

try on [TS PLAYGROUND](https://www.typescriptlang.org/play?filetype=d.ts#code/CYUwxgNghgTiAEBbA9sArhBByAzgNwHMA6AKxy3gG8AoeeSKHHeAEWUStrvuQDscALjDRgByGAAoAlFzpRgwCSEyIQvAQC54AURVqBUrQIAWASxxcAvtS4MmOves7d484AGFoTCbyiqtgjCmvASG8CbmVjZ0APQxPIgovEggJqjMAGbi8ABiMFAEqk4AZPDufAJQwSAwXMECNRlQYAjl6lW+AEaYztwNgtJavGiInTVRtl7MeQVFAvAgAB4NvMDMbBymiAAOjgLMbZXBUN0IlNaTjAcVHTULy2prDiBz8Fu7L-rX7cenvXT9ASDeDDUZ3OIgkAgYDwUAZaquAgdWTwABeyHYwNBY1qdGs1iAA)

we125182:
I use the dynamic extends to avoid dupliacted declarations. But the temp varibles(ContainableDom, ContainableElement) always export. which means user can import them, actually they didn't exist.  I add comment to explain. Maybe there has a good way.

Fuzzyma:
I don't think anyone will ever import them.
If everything else works like intended I would consider this a good solution :).
Does everything work?

we125182:
Yes, I linked the changed package in a project. `Fragment` has methods of `Containable` and `Dom`. `Container` keep same behavior.
image

Fuzzyma:
Awesome. Thank you!

we125182:
Glad to help.

PR #1321, opened 2024-07-17
fix: shape.css() typescript tip

hxdyj:
shape.css({
  backgroundColor:'red'   // ts success, but style set inoperative
})

shape.css({
  'background-color':'red'   // ts error, but style set success
})

this request fix typescript type  CSSStyleDeclaration

### DIFF
--- a/svg.js.d.ts
+++ b/svg.js.d.ts
@@ -4,26 +4,27 @@
 // trick to keep reference to Array build-in type
 declare class BuiltInArray<T> extends Array<T> {}
 
-// trick to have nice attribute list for CSS
-declare type CSSStyleName = Exclude<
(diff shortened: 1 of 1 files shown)


Fuzzyma:
Good catch! I think we also need to change the `CSSStyleName` parameters e.g. in `css<T extends CSSStyleName>(style: T): CSSStyleDeclarationWithVars[T]`. Otherwise the types for the css getter are broken.

hxdyj:
> 1f5f978

It my leak. fix [1f5f978](https://github.com/svgdotjs/svg.js/pull/1321/commits/1f5f978accc559f54aed0868e1a7a575a14a6d74)

Fuzzyma:
awesome thanks!

PR #1334, opened 2025-02-21
Remove gbox() from G in svg.js.d.ts

kcooney:
The method has not existed since c37d94a.


Fuzzyma:
ups :D - wow you really dug that commit up

PR #1341, opened 2025-09-18
Fix text.js in weird browsers

iapy:


Fuzzyma:
Please explain your changes, what issue they address and what implications they have

src/elements/Text.js
```diff
@@ -49,21 +49,21 @@ export default class Text extends Shape {
       const leading = this.dom.leading
 
       this.each(function (i) {
```
iapy, line 51: Within this anonymous function `this` is referencing function itself, not the Text instance as desired.

On line 47 reference to this is stored as self which I believe meant to be used.

src/elements/Tspan.js
```diff
@@ -49,7 +49,7 @@ export default class Tspan extends Shape {
     const dy = text.dom.leading * new SVGNumber(fontSize)
 
     // apply new position
-    return this.dy(i ? dy : 0).attr('x', text.x())
+    return this.dy(i ? dy : 0).attr('dx', text.x())
```
iapy, line 52: I'm not quite sure about this one, but it seems that when text is moved with `text.move(50, 50)`, the generated SVG fragment would look like

```
<text x="50" y="50">
<tspan dy="0" x="0"></tspan>
</text>
```

Which places tspan at x=0 in SVG coordinates. It has to be:
```
<text x="50" y="50">
<tspan dy="0" dx="0"></tspan>
</text>
```

iapy (commented):
I've left inline comments. Thanks.

src/elements/Text.js
Fuzzyma: this is referencing each child that is looped through. so every tspan. its not referencing the text

Fuzzyma:
@iapy both your comments are plain wrong. What bug is this PR solving?

src/elements/Tspan.js
Fuzzyma: Its correct as it is. A newline should start at x=0. Are you just guessing?

